Issue #17, opened 2021-01-11
[Task] Implement SQL schema for the tasks api


farhadbahrehmandhenry:
Task from Sprint 1 doc: Team 1, task 6.

farhadbahrehmandhenry:
this is to create database schema that will serve the app's API. 

# Azure MS SQL Database
create an Azure database with these specifications:
Azure SQL -> Single Database -> Server-less with 0.5 CPU min and 1 CPU max and Auto-pause after 1 h

# there are at least two ways we can run queries in the database

## using Azure Data Studio
1- download the Azure Data Studio from [here](https://docs.microsoft.com/en-us/sql/azure-data-studio/download-azure-data-studio?view=sql-server-ver15)
2- go to azure portal > your database resource group > click "connect with" 
![s1](https://user-images.githubusercontent.com/43734894/104536526-155cec00-55cd-11eb-9947-47e99595776f.png)

this will open a login form in Azure Data Studio that you just installed. you need to use the credential that used to create the server while creating the Azure MS SQL Database.
![login](https://user-images.githubusercontent.com/43734894/104543102-a2a63d80-55d9-11eb-9938-36d4b7e61cbd.png)

you may need to require to set a firewall rule. 
![firewall](https://user-images.githubusercontent.com/43734894/104543131-b6ea3a80-55d9-11eb-93be-8d61a1a08ce6.png)

in this case you can add a firewall rule for your database in Azure portal, by going to your database resource group, click on the database server, Firewalls and virtual networks, add the rule, and save.
![s2](https://user-images.githubusercontent.com/43734894/104543468-57d8f580-55da-11eb-85a4-ab3ddd995e66.png)

after adding the firewall you should be able to connect to your database, and start writing your queries.
![login-successfully](https://user-images.githubusercontent.com/43734894/104543527-750dc400-55da-11eb-8861-fb122f2ac811.png)

## using Query Editor in Azure portal
in this way you won't need installing Azure Data Studio. just go to the Azure portal, your database resource group, find the Query Editor, click on it, use the  credential that used to create the server while creating the Azure MS SQL Database, and then you can see your database and also write your queries.
![query](https://user-images.githubusercontent.com/43734894/104543556-848d0d00-55da-11eb-9b5e-bc643006537c.png)
![s3](https://user-images.githubusercontent.com/43734894/104543611-abe3da00-55da-11eb-8625-4ba19146b4e6.png)



## what data I added to database for this project
[...]

PR #29, opened 2021-01-14
Feature/team1 task6 issue17 implement sql

farhadbahrehmandhenry:
issue #17


farhadbahrehmandhenry:
this is to create database schema that will serve the app's API. 

# Azure MS SQL Database
create an Azure database with these specifications:
Azure SQL -> Single Database -> Server-less with 0.5 CPU min and 1 CPU max and Auto-pause after 1 h

# there are at least two ways we can run queries in the database

## using Azure Data Studio
1- download the Azure Data Studio from [here](https://docs.microsoft.com/en-us/sql/azure-data-studio/download-azure-data-studio?view=sql-server-ver15)
2- go to azure portal > your database resource group > click "connect with" 
s1

this will open a login form in Azure Data Studio that you just installed. you need to use the credential that used to create the server while creating the Azure MS SQL Database.
login

you may need to require to set a firewall rule. 
firewall

in this case you can add a firewall rule for your database in Azure portal, by going to your database resource group, click on the database server, Firewalls and virtual networks, add the rule, and save.
s2

after adding the firewall you should be able to connect to your database, and start writing your queries.
login-successfully

## using Query Editor in Azure portal
in this way you won't need installing Azure Data Studio. just go to the Azure portal, your database resource group, find the Query Editor, click on it, use the  credential that used to create the server while creating the Azure MS SQL Database, and then you can see your database and also write your queries.
query


## what data I added to database for this project
1- users table, including 22 users data => user id, first name, last name, email, deleted
2- tasks table including 15 tasks data => task id, task, created time, deleted
3- relation between users and tasks is a many to many relation so we have a join table including 25 records  => task id, user id, assigned date, accomplished

sql/dbschema.sql
```diff
@@ -0,0 +1,99 @@
...
+  task VARCHAR(300),
+  createdTime DATETIME NOT NULL,
+  deleted INT NOT NULL DEFAULT 0
+);
+
+CREATE TABLE userTasks(
+  taskId INT NOT NULL,
+  userId INT NOT NULL,
+  assignedDate DATETIME NOT NULL,
+  accomplished INT NOT NULL DEFAULT 0,
```
toddysm, line 20: `completed` may be more appropriate name for the field.

sql/dbschema.sql
```diff
@@ -0,0 +1,99 @@
...
+  
+  CONSTRAINT FK_Task_User FOREIGN KEY (taskId)
+    REFERENCES  tasks(taskId)
+    ON DELETE CASCADE,
+  CONSTRAINT FK_User_Task FOREIGN KEY (userId)
+    REFERENCES  users(userId)
+    ON DELETE CASCADE
+);
+
+INSERT INTO  users (firstName, lastName, email, deleted)
```
toddysm, line 30: In general, you should not populate the database with sample data.

sql/dbschema.sql
```diff
@@ -0,0 +1,99 @@
...
+  userId INT PRIMARY KEY IDENTITY(1,1),
+  firstName VARCHAR(50) NOT NULL,
+  lastName VARCHAR(50) NOT NULL,
+  email VARCHAR(50) NOT NULL,
+  deleted INT DEFAULT 0 NOT NULL
+);
+
+CREATE TABLE tasks(
+  taskId INT PRIMARY KEY IDENTITY(1,1),
+  task VARCHAR(300),
```
toddysm, line 11: What does `task` mean in this context? Is this the task's `title` or `description`? Maybe having two fields - one for short title and another for long description is more appropriate.

sql/dbschema.sql
```diff
@@ -0,0 +1,99 @@
+CREATE TABLE users (
+  userId INT PRIMARY KEY IDENTITY(1,1),
+  firstName VARCHAR(50) NOT NULL,
+  lastName VARCHAR(50) NOT NULL,
+  email VARCHAR(50) NOT NULL,
+  deleted INT DEFAULT 0 NOT NULL
```
toddysm, line 6: Why do we need `deleted`? What is the purpose of having `deleted` field?

sql/dbschema.sql
```diff
@@ -0,0 +1,99 @@
...
+);
+
+CREATE TABLE tasks(
+  taskId INT PRIMARY KEY IDENTITY(1,1),
+  task VARCHAR(300),
+  createdTime DATETIME NOT NULL,
+  deleted INT NOT NULL DEFAULT 0
+);
+
+CREATE TABLE userTasks(
```
toddysm, line 16: Why do we need a separate table linking the users and the tasks? Is the relationship user:task 1:1, 1:* or *:*?

toddysm (changes requested):

sql/dbschema.sql
farhadbahrehmandhenry: insert statements removed from code. sample data in database will be removed as soon as we test the api functions.

sql/dbschema.sql
farhadbahrehmandhenry: I thought it might be a safer solution to update deleted field from 0 to 1 when a user or task gets deleted. in this way we avoid delete statement for more security and also we can restore the deleted record if we want. in this case the select statement will be like this => 
SELECT * FROM users WHERE deleted = 0;

I can remove it if you think it's redundant.

sql/dbschema.sql
farhadbahrehmandhenry: I think it's a many-many relation. one user can have several tasks, and one task can be assigned to multiple users.
I can make it one-one if you think many-many is not necessary here.

theKunte: I would agree with Fahrrad, a user can have many tasks and the task can have multiple users. *:*
But we can also structure it that only a user can be assigned to different tasks 1:*

theKunte: After talking with the team we decided a 1:* relationship is appropriated compare to the *:* relationship

theKunte (changes requested):
After talking with the team we decided a 1:* relationship is appropriated compare to the: relationship

theKunte (commented):
The relationship with the task and user git solved. I can see that you removed the many to many relationships. Now we have one to many. I was able to connect to your DB with the function userId and pull the sample data. As for now, I would approve this branch.

theKunte (approved):

sql/dbschema.sql
```diff
@@ -0,0 +1,26 @@
...
+  userId INT NOT NULL,
+  title VARCHAR(50),
+  description VARCHAR(400),
+  createdTime DATETIME NOT NULL,
+  deleted INT NOT NULL DEFAULT 0,
+  completed INT NOT NULL DEFAULT 0
+
+  CONSTRAINT FK_User_Task FOREIGN KEY (userId)
+    REFERENCES  users(userId)
+    ON DELETE CASCADE
```
lbrown51, line 20: This probably makes sense to include in the tasks table setup, but if we never actually delete users would this ever come into play?

sql/dbschema.sql
```diff
@@ -0,0 +1,26 @@
...
+  email VARCHAR(50) NOT NULL,
+  deleted INT DEFAULT 0 NOT NULL
+);
+
+CREATE TABLE tasks(
+  taskId INT PRIMARY KEY IDENTITY(1,1),
+  userId INT NOT NULL,
+  title VARCHAR(50),
+  description VARCHAR(400),
+  createdTime DATETIME NOT NULL,
```
lbrown51, line 14: Do we also want to include when the task was completed?

lbrown51: Further, do we want to include a due date for tasks? That seems like it'd be pretty useful. I know that having a perfect task management app is not exactly the point of this project, but I still think it might be worth thinking about?

sql/dbschema.sql
lbrown51: Do we want to have a created at field as well?

farhadbahrehmandhenry: good idea, I will add it.

sql/dbschema.sql
farhadbahrehmandhenry: no, but if we decide to delete a user in the future, we won't be worried about modifying the tasks table. do you think it's not necessary here?

sql/dbschema.sql
farhadbahrehmandhenry: good ideas, I will add those.

sql/dbschema.sql
nataliagourova: From the perspective of someone who writes the sql query based on the db setup, "created" and "deleted" make logical sense, but will not be returned as part of API response. With that, I'd vote to leave out

sql/dbschema.sql
nataliagourova: I think this is a terrific idea for maybe next sprint, but not for sprint 1 with the time we have left. For example, as I am writing the script to "update" a task, I am currently focusing on the existing fields, and would even forgo the create timestamp at this point of time (scope creep).

sql/dbschema.sql
nataliagourova: I'd say the simpler the better this time around, so that we do not run into 500 internal server errors when we delete stuff and cannot.

sql/dbschema.sql
nataliagourova: I like having the two tables and not having to introduce the reference table, it simplifies queries for us and the teams who pick this up in following sprints.

sql/dbschema.sql
```diff
@@ -8,26 +8,19 @@ CREATE TABLE users (
...
+  userId INT NOT NULL,
   title VARCHAR(50),
   description VARCHAR(400),
   createdTime DATETIME NOT NULL,
-  deleted INT NOT NULL DEFAULT 0
-);
+  deleted INT NOT NULL DEFAULT 0,
+  completed INT NOT NULL DEFAULT 0
 
-CREATE TABLE userTasks(
```
nataliagourova: Love that you dropped this.

nataliagourova (approved):
I left some inline comments, overall looks good!
Was the script to populate tables added after the third table drop? I could not locate it.

toddysm (approved):

PR #40, opened 2021-01-15
Feature/toddys examples

toddysm:
Examples of the automation scripts and the environment variables for DB access


example/functions-example/HttpTriggerToddysmExample/function.json
```diff
@@ -0,0 +1,20 @@
+{
+  "scriptFile": "__init__.py",
+  "bindings": [
+    {
+      "authLevel": "anonymous",
+      "type": "httpTrigger",
+      "direction": "in",
+      "name": "req",
+      "methods": [
```
nataliagourova, line 9: adding "delete" would be helpful

nataliagourova (approved):

example/functions-example/HttpTriggerToddysmExample/function.json
theKunte: would we add patch? Just a question in regards to last quarter for the different responses.

toddysm: @nataliagourova and @theKunte depends on what endpoint this function implements. If it is only the `/users` you need to think whether `PUT/PATCH` and `DELETE` are appropriate.

example/functions-example/HttpTriggerToddysmExample/__init__.py
```diff
@@ -0,0 +1,28 @@
+import logging
+import os
+
+import azure.functions as func
+
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
+    logging.info('Python HTTP trigger function processed a request.')
+
+    db_username = os.environ["ENV_DATABASE_USERNAME"]
```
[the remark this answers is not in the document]
toddysm: Not sure what you mean with *"we didn't actually run the endpoint url in the demo to see the result"* - can you clarify?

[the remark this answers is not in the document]
farhadbahrehmandhenry: double checked the lecture, I was wrong. however I was able to access the env variables by putting them under "values" in local.settings.json and getting them by importing "os" and calling os.environ[], exactly like what you had done in the lecture. just wanted to confirm it's working properly.

farhadbahrehmandhenry (approved):

example/automation-example/function-automation/deploy-template.ps1
```diff
@@ -0,0 +1,21 @@
...
+    [Parameter(Mandatory=$True)]
+    [string]
+    $ServicePrincipalId,
+   
+    [Parameter(Mandatory=$True)]
+    [string]
+    $ServicePrincipalPassword
+)
+
+Connect-AzAccount
```
toddysm, line 19: Can you change something here?

farhadbahrehmandhenry: I believe this is good to merge!

Issue #41, opened 2021-01-16
[TASK] Implement script to create a public IP address


natni:

natni:
Is your feature request related to a problem? Please describe.
N/A

Describe the solution you'd like
I want to create a script that creates Public IP from the user input

Describe alternatives you've considered
N/A

Additional context
N/A

PR #43, opened 2021-01-16
A script that creates public ip address.

natni:


example/Automation/create_ip_address.ps1
```diff
@@ -0,0 +1,17 @@
...
+# Location of Azure resources
+$loc = 'westus2'
+
+# Name of the azure public address
+$pubIP = 'public-ip-standard-nsc-ad440'
+
+# Make the public IP a standard zonal public IP address in Zone 2
+$sku = 'Standard'
+$alloc = 'Static'
+$zone = 2
```
toddysm, line 15: What does `$zone=2` mean? Comments in your code will also be helpful.

toddysm (changes requested):
Also, assign people as reviewers

natni:
Thanks for the feedback. I have updated the script based on the recommendations.

natni:
**Description**
Closes #41 

**This PR holds a script that creates a Public IP Address in Azure** 
You should be careful about what you enter on the prompts when creating Public IP. Enter a correct resource group that exists, a proper location eg. 'eastus2', a proper public address name, SKU type eg. 'Standard' and a proper allocation method eg 'static' 

How to test:
1. Run the script while on the file directory by writing _./create_ip_address.ps1_
2. Enter inputs for the params: _rgname_, _location_, _publicIp_, _sku_ and _allocation_.

Automation/create_ip_address.ps1
```diff
@@ -0,0 +1,13 @@
...
+param(
+        [string] [Parameter(Mandatory=$true)] $rgname, # Name of the resource group
+        [string] [Parameter(Mandatory=$true)] $location, # Location of Azure resources
+        [string] [Parameter(Mandatory=$true)] $publicIp, # Name of the azure public address
+        [string] [Parameter(Mandatory=$true)] $sku, # Type of SKU, Standard or Basic
+        [string] [Parameter(Mandatory=$true)] $allocation # AllocationMethod, Static or Dynamic
+      )
+
+# Creates a new Public IP Address based on the params.
+New-AzPublicIpAddress -ResourceGroupName $rgname -Name $publicIp -Location $location -AllocationMethod $allocation -SKU $sku
```
unicornsleuth, line 13: We don't want there to be any user interaction/prompting, so the directions should say to add all of it on the same line, like 
`./Automation/create_ip_address.ps1 -rgname test-natanaels-pr-ly -location "uksouth" -publicIp "ip-address-test-ly" -sku Basic -allocation Dynamic`

Also consider doing some checks on the resource group, have it create a new rg if that one doesn't already exist
You will get errors if you try to create an IP address with a location that doesn't match the RG, so you can use the same location for both.

unicornsleuth (changes requested):

Automation/create_ip_address.ps1
```diff
@@ -0,0 +1,30 @@
+## Command to create a public ip ##
```
unicornsleuth, line 1: Do we need this file? Not sure which is the final/correct version of the script

Automation/public_ip_template.json
```diff
@@ -0,0 +1,29 @@
+{
```
unicornsleuth, line 1: Same with this - pretty sure just needs to be deleted

unicornsleuth (changes requested):

natni:
For some reason, I didn't save my code on VS Code before pushing and it just pushed stuff I already modified and deleted. I'm pushing a new one now

Automation/PublicIP/public_ip_template.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "string"
+        }
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.Network/publicIPAddresses",
+            "apiVersion": "2020-05-01",
+            "name": "[parameters('pubIpName')]",
+            "location": "westus2",
```
unicornsleuth, line 20: Shouldn't this be [parameters('location')]  ?

Automation/PublicIP/public_ip_template.json
```diff
@@ -0,0 +1,33 @@
...
+            "sku": {
+                "name": "Standard"
+            },
+            "properties": {
+                "ipAddress": "20.69.187.53",
```
unicornsleuth, line 25: hard-coded value?

Automation/PublicIP/create_ip_address.ps1
```diff
@@ -0,0 +1,32 @@
+## Command to create a public ip ##
+
+# Take parameter used to create Public IP Address
+param(
+        [string] [Parameter(Mandatory=$true)] $tenantId, # Tenant ID found in the Key Vault     
+        [string] [Parameter(Mandatory=$true)] $applicationId, # Application ID found in the Key Vault  
+        [string] [Parameter(Mandatory=$true)] $secret, # Secret Key found in the Key Vault            
+        [string] [Parameter(Mandatory=$true)] $subscriptionId, # Subscription ID found in the Key Vault  
+        [string] [Parameter(Mandatory=$true)] $rgname, # Name of the resource group
```
unicornsleuth, line 9: I think Toddy said to use longer, descriptive variable names, so he'd prefer resourceGroupName. We should check if he wants camel case or all the first letters capitalized, but low priority

unicornsleuth (commented):
It successfully created the IP address! Looking great - just a couple small comments and then you're good!

unicornsleuth (approved):
Everything looks good and works as expected! When I first tried, I put in a new resource group and it didn't create anything, but the error was handled smoothly, and when I selected a rg that already existed, it worked perfectly! : )

toddysm (approved):

PR #44, opened 2021-01-17
Sg artillery perf Test

gillilands19:
**What is It:**

An Artillery YAML script to load test the given URL. Ramps up to 10 users per second for a total of 1 hour of load testing. Includes a custom logging function for responses.
or responses.

**Notes**

It appears that Artillery does not support variables in query strings. Harcoding a query string 
e.g https://nsc-fun-dev-usw2-tsmtestfunction.azurewebsites.net/api/HttpTriggerLoadTest?name=Sean
also returns an invalid url error during the run. When looping through variables for a query string the response times section also returns NaN for all metrics.

For now I have set this up to just make request to the base url with no query strings. I've also included a custom JS module with a function that logs each response code for each Virtual User that makes a request. You can run the script and have it create a Json output file with the results of the test. It can get very long and detailed, so for performance reasons we may want to find another solution for running an hour long test. For example a 10 minute test generated nearly 17,000 lines of json output. You can then take that json and generate a default html report using the Artillery CLI. 

We may also want to utilize the artillery-plugin-expect package for writing more meaningful assertions and expectations.

**Installing Artillery**

You must have node.js and npm installed. In terminal install artillery globally:
`npm install -g artillery`

**Running the script:**

1. Change directory to test/artillery
2. Type the following command in your terminal:
`artillery run perf-test-artillery.yaml`

If you want the json output of the results:
`artillery run --output report.json perf-test-artillery.yaml`

**Time Spent:**

- 2 hours reading artillery documentation
- 2 hours testing different scenarios and features and reading github issues.
- 30 minutes finalizing yaml script and writing JS logging function

**Links:**

https://artillery.io/docs/guides/getting-started/core-concepts.html#High-level-view

https://github.com/artilleryio/artillery/issues/698


jaguiniga (approved):
The code looks good. Other than the git ignore conflict, which I think I saw Toddy mention about something on the Slack. But other than that I think that was the only merge conflict that is popping up. I see the config file that is testing the URL and the scenarios like load test. Also the function and the log response as the output after the tests

test/artillery/report.json
```diff
@@ -0,0 +1,3702 @@
+{
```
zxkevinxz, line 1: I believe this could be added to the gitignore.

zxkevinxz (changes requested):
I tested it, and it worked great. It looks like you did your research on using variables in the script as well. I just had 1 small comment.

gillilands19:
I've completed the requested changes and fixed merge conflicts. Thanks.

test/artillery/report.json
gillilands19: Done and Done. Merge conflicts are also fixed.

zxkevinxz (approved):
Changes look good.

PR #47, opened 2021-01-18
Yaholl/issue 16 vnet script

### LINKED ISSUE #16: [TASK] Implement script to create Virtual Network
Is your feature request related to a problem? Please describe.
Providing automation for the creation of a resource in Azure

Describe the solution you'd like
A script with a JSON file that can provision a virtual network automatically

Describe alternatives you've considered
n/a

Additional context
n/a

unicornsleuth:
closes #16 

*What this PR Does*
- creates a parameter file for a vnet (currently without subnets)
- creates a script for logging in, which can be called before the other scripts
- creates a script for creating a virtual network with a specified name (mandatory) and address prefix (optional)

*To Test*
- navigate to the repo/Automation/vnet and use the following command:
- `./create_vnet.ps1 -tenantId {tenant id} -subscriptionId {subscription id} -applicationId {application id} -secret {secret} -resourceGroupName {rg name} -location {location} -vNetName {vnet name}`


Automation/create_vnet_script.ps1
```diff
@@ -0,0 +1,26 @@
+# to run this, navigate to the repo and run ./Automation/create_vnet_script.ps1
```
jleeboyd, line 1: I think it may make sense to put these files into a separate Vnet folder within the Automation folder.

Automation/vnet_template.json
```diff
@@ -0,0 +1,45 @@
...
+    "location": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().location]",
+      "metadata": {
+        "description": "Location for all resources."
+      }
+    },
+    "utcValue": {
+      "type": "string",
+      "defaultValue": "[utcNow()]"
```
jleeboyd, line 28: Is this 'utcValue' parameter required for this vnet template, or for use in a later PR?

Automation/login.ps1
```diff
@@ -0,0 +1,14 @@
...
+        [string] [Parameter(Mandatory=$true)] $subscriptionId
+      )
+
+[securestring]$secureSecret = ConvertTo-SecureString $secret -AsPlainText -Force      
+[pscredential]$credObject = New-Object System.Management.Automation.PSCredential ($applicationId, $secureSecret)
+Connect-AzAccount -ServicePrincipal -Credential $credObject -Tenant $tenantId      
+write-output "Logged into the account $applicationId"
+
+Set-AzContext -Subscription $subscriptionId
+write-output "Set to subscription $subscriptionId"
```
jleeboyd, line 14: Is the subscriptionId required for context when logging in with a service principal?

Automation/create_vnet_script.ps1
```diff
@@ -0,0 +1,26 @@
...
+        [string] [Parameter(Mandatory=$true)] $secret,
+        [string] [Parameter(Mandatory=$true)] $subscriptionId,
+        [string] [Parameter(Mandatory=$true)] $resourceGroupName,
+        [string] [Parameter(Mandatory=$true)] $location,
+        [string] [Parameter(Mandatory=$true)] $vNetName,
+        [string] [Parameter(Mandatory=$false)] $vNetAddressPrefix
+      )
+
+$pathToVNetTemplate = "./Automation/vnet_template.json"    
+# $pathToVNetTemplate = "D:\VSCode\ad440-winter2021-thursday-repo\Automation\vnet_template.json"
```
jleeboyd, line 16: I believe this path name has an extra 'Automation'.

jleeboyd (changes requested):
When initially testing I ran into this error, due to the template path name (see code comment). However after changing and re-running the script it created the vnet + resource group successfully, nice work!

Automation/create_vnet_script.ps1
unicornsleuth: Yeah I saw you do that for the sg, I thought it looked like a good idea - we should all agree on the same structure though

Automation/vnet_template.json
unicornsleuth: tbh I could remove it, it was part of the template I used. I don't think it needs to be there.

Automation/login.ps1
unicornsleuth: No, it was just to set the subscription/context, since I think it makes sense to do that at the same time as logging in.

Automation/create_vnet_script.ps1
unicornsleuth: You're right! Good catch - I need to fix this

natni: The script looks great but I believe some comments for the params would help. Like what tenantId and secrete are about.

unicornsleuth: Fixed - I added comments for the applicationId and the secret, but I'm not sure what to say about the tenantId - it's just the tenantId associated with the account, I think.

Automation/vnet_template.json
unicornsleuth: Done!

zxkevinxz (approved):
I tested it and went over all of the code with you. All the changes and everything is really well thought out. All the values are params and nothing seems to be hardcoded.

natni (approved):
I really like the new change. Works great! Maybe some comments for the params found in the create_vnet.ps1 file might help.

jleeboyd (approved):

PR #65, opened 2021-01-21
pull request due to sprint 1

jinudean:
pull request for testing and reviewing purpose


jinudean:
first pull request due to sprint 1

Automation/createresource.ps1
```diff
@@ -0,0 +1,12 @@
+$resourceGroupName = Read-Host -Prompt "Enter the Resource Group name"
+$location = Read-Host -Prompt "Enter the location (i.e. centralus)"
+$storageAccountName = Read-Host -Prompt "Enter the storage account name"
```
zxkevinxz, line 3: This could be better if taken in as parameters instead of prompts so whoever runs it can just do it in one line.

Automation/createresource.ps1
```diff
@@ -0,0 +1,12 @@
...
+
+# Create the storage account.
+$storageAccount = New-AzStorageAccount -ResourceGroupName $resourceGroupName `
+  -Name $storageAccountName `
+  -Location $location `
+  -SkuName "Standard_LRS"
```
zxkevinxz, line 9: Should this SkuName be hardcoded? Maybe a parameter with a default value?

zxkevinxz (changes requested):

PR #81, opened 2021-01-27
Complete implementation of Azure function for /users endpoint - Task 76

### LINKED ISSUE #76: [Task] Implement logging and complete functionality for Azure function /users
The problem is that the current functionality of the Azure function app /users only does get requests. Further there is no logging.

Implement those two functionalities. 

These are core functionalities and there are not really any alternatives.

lbrown51:
Resolves #76 
- This code completes the implementation of the Azure function for the /users endpoint
- The function will be deployed automatically when merged into the development branch
- It doesn't require configuration to be run on the cloud
- If you want to test it locally
  - Make sure you have the Azure functions extension for VSCode
  - Pull the latest iteration of this branch
  - Install the [ODBC driver](https://docs.microsoft.com/en-us/sql/connect/odbc/download-odbc-driver-for-sql-server?view=sql-server-ver15) necessary to connect to the database
  - In local.settings.json at the root of the functions folder, enter in the values for the keys below in the following format: { "Values": { "KEY": "VALUE" } }
    - ENV_DATABASE_NAME (name of the database)
    - ENV_DATABASE_PASSWORD
    - ENV_DATABASE_SERVER (this is the server location)
    - ENV_DATABASE_USERNAME
  - You can click F5 or click run->Start Debugging in the functions folder to start it up
- The implemented function only makes sense for GET and POST requests
  - The GET request doesn't take anything and returns all users
  - The POST request takes a JSON object { "firstName": "a first name", "lastName": "a last name", "email": "an@email.com" } which returns the id of the newly created user in form { "userId": 5 }
  - To test using JSON, copy the link provided by the debugger (the "/users" one) into Postman, and use as usual for GET and POST (info goes into the body)
- All other troubleshooting problems that I know about should be taken care of by error handling but let me know if you run into one.


nataliagourova:
Why would deleting all users not be an option/would not make sense? If no delete method is implemented, we probably do not need it listed in function.json

lbrown51:
@nataliagourova In order of comment

1. We talked last quarter in class about how a delete to /users gets rid of every single user in the database, which you wouldn't ever really do via REST API because it's such a huge action, which is why I didn't include it.
2. I included delete and put (even though I didn't implement them) in the bindings because I want all the standard REST commands to be callable even if they are not implemented. I'm not sure that that's correct, @toddysm?
3. I didn't do anything in regards to venv environments so I'm not really sure what I'd put there. Is that just for installing py3.8 + azure-functions + pyodbc + pypyodbc?
4. I'll clarify how to pass the json object! Good call!

functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -1,52 +1,136 @@
...
+            logging.debug("User added successfully!")
+            return new_user_id_http_response
+
+        else:
+            logging.warn(f"Request with method {method} has been recieved, but that is not allowed for this endpoint")
+            return func.HttpResponse(status_code=405)
+
+    #displays erros encountered when API methods were called
+    except Exception as e:
+        return func.HttpResponse("Error: %s" % str(e), status_code=500)
```
nataliagourova, line 49: Would logging.critical be more appropriate here?

functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -1,52 +1,136 @@
...
+        logging.critical('No method available')
+        raise Exception('No method passed')
+
+    # Create a new connection
+    logging.debug("Attempting DB connection!")
+    try:
+        conn = get_db_connection()
+    except (pyodbc.DatabaseError, pyodbc.InterfaceError) as e:
+        logging.error("Failed to connect to DB: " + e.args[0])
+        logging.debug("Error: " + e.args[1])
```
nataliagourova, line 23: Would logging.critical work better for 500 status codes?

nataliagourova (approved):

functions/HttpTriggerAPIUsers/__init__.py
lbrown51: Good question! I think it's kind of a toss up. According to Toddy's article on log levels, the critical level is when the application cannot recover and must exit, whereas the error level is when the condition is noticeable by the user and impacts the normal execution of the application. After reading a bit more, I _think_ that error might be the correct level because the function can still be called again and hasn't really crashed, **BUT** I'm totally open to alternative opinions on the matter.

lbrown51: I'm not exactly sure, but I don't think it being a 500 has a ton of bearing on the log level, but I could be wrong.

functions/HttpTriggerAPIUsers/__init__.py
lbrown51: I think probably the same case as above

functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -1,52 +1,136 @@
...
+                         OUTPUT inserted.userId INTO @NEWID(ID)
+                         VALUES('{}', '{}', '{}');
+
+                         SELECT ID FROM @NEWID
+                         """.format(firstName, lastName, email)
+
+
+        logging.debug(
+            "Using connection cursor to execute query (add a new user and get id)")
+        count = cursor.execute(add_user_query)
```
farhadbahrehmandhenry, line 129: [this](https://stackoverflow.com/questions/43491381/pyodbc-the-sql-contains-0-parameter-markers-but-1-parameters-were-supplied-hy0) might be useful in order to avoid injection.

farhadbahrehmandhenry (approved):

functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -1,52 +1,136 @@
...
+    method = req.method
+    if not method:
+        logging.critical('No method available')
+        raise Exception('No method passed')
+
+    # Create a new connection
+    logging.debug("Attempting DB connection!")
+    try:
+        conn = get_db_connection()
+    except (pyodbc.DatabaseError, pyodbc.InterfaceError) as e:
```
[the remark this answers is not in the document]
lbrown51: That's a good point. It seems like pyodbc has been updated more recently, so maybe we should change the other two to use it?

functions/HttpTriggerAPIUsers/__init__.py
lbrown51: I also saw that syntax, but didn't realize it could help with injection attacks! Good catch!

functions/HttpTriggerAPIUsers/__init__.py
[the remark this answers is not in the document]
theKunte: I think some people had issues using pyodbc on their machines, but by using pypyodbc the issue got resolved because it can be used cross-platform.

functions/HttpTriggerAPIUsers/__init__.py
theKunte: In the example with the connection to the DB and the 500 status code, I would agree with Natalia to change it to critical since the application can not continue.  logging.critical would be more appropriate for this 500 status code

functions/HttpTriggerAPIUsers/__init__.py
theKunte: I am leaning more towards logging.critical

theKunte (approved):
Good work. The only thing that I would change is the logging.error to logging.critical. Everything else looks good. Good comments! I was able to test everything with the instruction.

functions/HttpTriggerAPIUsers/__init__.py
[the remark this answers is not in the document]
lbrown51: I'm gonna leave switching over to another issue later

nataliagourova:
one last thing: my linter is complaining  about an unused variable in your code on line 129:
{
	"owner": "python",
	"code": "unused-variable",
	"severity": 4,
	"message": "Unused variable 'count'",
	"source": "pylint",
	"startLineNumber": 129,
	"startColumn": 9,
	"endLineNumber": 129,
	"endColumn": 9
}

nataliagourova (commented):
Potential for sql injection

functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -1,52 +1,137 @@
...
-          return func.HttpResponse(
-              "Unauthorized",
-              status_code=403
-          )
-
-# Define query    
-def select_users_query(conn):
+    connection_string = "Driver={};Server={};Database={};Uid={};Pwd={};Encrypt=yes;TrustServerCertificate=yes;Connection Timeout=30;".format(
+        driver, server, database, username, password)
+
```
nataliagourova, line 68: Potential for sql injection

lbrown51: I don't know that that's true. These variables come from our environment variables so it's not really something a user has access to. Also, I'm not sure that ? would work because we aren't executing a statement.

lbrown51:
Thanks for this! I also changed my code while listening to Toddy's comments, so I should have everything. I'll make a new PR later.

PR #88, opened 2021-01-28
[BUG] users/userId/tasks/taskId API logging and error handling Issue #83

### LINKED ISSUE #83: [BUG] JSON output, logging and method updates  for /users/{user_id}/task/{task_id} API
**Describe the bug**
The taskId/userId Https trigger function is deployed to the cloud and is functional.
The function does not produce a JSON output, has limited logging implemented and does not have a fully implemented DELETE method

**To Reproduce**
run the function per the test instructions and observe the output not being in JSON format, with limited logging and  incomplete DELETE method functionality

**Expected behavior**
When GET, UPDATE and DELETE methods are run, the function is expected to produce JSON output, and robust logging info should be available

**Screenshots**
see the testing instructions with non-conformant output in this PR #61 

**Additional context**
tbd

nataliagourova:
### Resolves #83 

This code completes the implementation of the Azure function for the `/users/user_id/tasks/task_id endpoint`
The function will be deployed to the cloud following the approval and merge into the development branch
It doesn't require configuration to be run on the cloud
### To test locally:
Make sure you have the Azure functions extension for VSCode
Pull the latest iteration of this branch
Install the ODBC driver necessary to connect to the database (pyodbc or pypyodbc)
Configure your venv and vscode files to set up the appropriate runtime for this Azure function
In local.settings.json at the root of the functions folder, enter in the values for the keys below in the following format: { "Values": { "KEY": "VALUE" } }
ENV_DATABASE_NAME (name of the database)
ENV_DATABASE_PASSWORD
ENV_DATABASE_SERVER (this is the server location)
ENV_DATABASE_USERNAME

In VSCode, click `F5 `or `run->Start Debugging `in the functions folder to fire up the function
After the function has started, input the url provided in terminal window into Postman, and test the functionality by selecting the appropriate method and url parameters and/or json body, like so:

**GET method**
Pass the userId and taskId as url parameters, JSON body is not required
Screen Shot 2021-01-31 at 5 20 07 PM

**PUT method**:
Pass the userId and taskId as url parameters, JSON body IS required to include `title`, `description`, `dueDate`, `completed`, `completedDate` (all five updatable fields need to be present, all other JSON fields will be ignored), you may pass more by cutting and pasting from GET results
You may use this sample JSON for update testing:
```
{
    "userId": 16,
    "taskId": 24,
    "title": "My Question",
    "description": "Decide on if to be or not to be",
    "dueDate": null,
    "completed": 1,
    "completedDate":"21/01/07 05:46:38"
}
```
Note that the in-built `datetime `conversion will only work for legitimate `yy/mm/dd hh:mm:ss` If you attempt to update the due date to occur on 20/13/32, you will receive an error.
**Be sure to add the data in the `"yy/mm/dd hh:mm:ss"` string format**, otherwise you will receive an error.

successful request results:
Screen Shot 2021-01-31 at 5 20 47 PM
**PATCH method**
Same instructions as for PUT, userId and taskId are required, JSON body may contain one to five key-value pairs excluding userId and taskId. If "Completed" is included, it may not be null per db specifications.
Screen Shot 2021-01-31 at 5 35 52 PM

PUT and PATCH results: 
Screen Shot 2021-01-31 at 5 36 19 PM

**DELETE method**
Pass the userId and taskId as url parameters, JSON body is not required.
_Please DO NOT delete the record with userId = 16, taskId = 24
taskId = 25-26 for userId =16 are available for the delete functionality testing.
If the requested taskId is not in the db, you will receive an error message:
Screen Shot 2021-01-28 at 2 44 17 PM

### DIFF
--- a/functions/HttpTriggerAPIUsersIdTaskId/__init__.py
+++ b/functions/HttpTriggerAPIUsersIdTaskId/__init__.py
@@ -10,7 +10,6 @@ import azure.functions as func
 #GET API method function
 def get(userId, taskId):
     #connects to db
-
     try:
         logging.debug('Attempting a db connection')
         cnxn = connect()
@@ -18,6 +17,7 @@ def get(userId, taskId):
         logging.debug('opened connection')        
         logging.debug(f'Attempting to execute GET task query for task {taskId}')
         #Get task title, description and user name by userId and taskId
+        #Avoids using SELECT * to prevent retuning unwanted unformation
         sql_query = ("""SELECT tasks.userId, CONCAT (users.firstName, ' ', users.lastName) AS "user",
                     tasks.taskId, tasks.title, tasks.description, tasks.createdDate, tasks.dueDate, 
                     tasks.completed, tasks.completedDate 
@@ -43,30 +43,80 @@ def get(userId, taskId):
 #PUT API method function
 def update(userId, taskId, task_fields):
     #connects to db
-    cnxn = connect()
-    cursor = cnxn.cursor()
     try:
+        cnxn = connect()
+        cursor = cnxn.cursor()
         logging.debug('opened connection')        
-        logging.debug(f'Going to execute UPDATE task {taskId} for user {userId} query')
-        # task_req_body input check
-        logging.debug("Checking the request body for necessary fields to update a task")    
-        # update task: with all passed JSON fields
+        logging.debug(f'Going to execute UPDATE query task {taskId} for user {userId}') 
+        try:
+            # update task: with 5 required JSON fields
+            assert len(task_fields) == 4, "Pass five required fields to update the task"
+        except AssertionError as req_body_content_error:
+            logging.error('Query did not contain all fields to update the task') 
+            return func.HttpResponse(req_body_content_error.args[0], status_code=400)
         sql_query = """UPDATE [dbo].[tasks]
             SET title = ?, description = ?, dueDate = ?, completed = ?, completedDate = ?
-            WHERE userId = ? AND taskId = ?"""
-        try:
-            rowcount = cursor.execute(sql_query, task_fields.get('title'), task_fields.get('description'), 
-            task_fields.get('dueDate'), task_fields.get('completed'), task_fields.get('completedDate'), 
-            userId, taskId).rowcount
-            if not rowcount:
-                logging.error('Invalid input and/or no record with the requested parameters exists in db')
-                return func.HttpResponse('Invalid input', status_code=404)                
-            logging.debug('UPDATE query executed')
-            logging.debug(f"Executed the query: {rowcount} rows affected for taskId {taskId}")
-            return func.HttpResponse(status_code=200)
-        except Exception as e:
-            logging.critical('Unable to execute the query')
-            return func.HttpResponse("Error: %s" % str(e), status_code=400)
+            WHERE userId = ? AND taskId = ?"""   
+
+        rowcount = cursor.execute(sql_query, task_fields.get('title'), task_fields.get('description'), 
+        task_fields.get('dueDate'), task_fields.get('completed'), task_fields.get('completedDate'), 
+        userId, taskId).rowcount
+        if not rowcount:
+            logging.error('Bad input and/or no record with the requested parameters exists in db')
+            return func.HttpResponse('Bad or invalid input', status_code=404)                
+        logging.debug('UPDATE task query executed')
+        logging.debug(f"Executed the query: {rowcount} rows affected for taskId {taskId}")
+        return func.HttpResponse(status_code=200)
+    except Exception as e:
+        logging.critical('Unable to execute the query')
+        return func.HttpResponse("Error: %s" % str(e), status_code=400)
+    finally:
+        #commits changes to db
+        cnxn.commit()
+        #properly closes the connection
+        cursor.close()
+        cnxn.close()
+        logging.debug('Closed the db connection')    
+
+#PATCH API method function, permits min 1 field passed in task_fields
+def patch(userId, taskId, task_fields):
+    #connects to db
+    cnxn = connect()
+    cursor = cnxn.cursor()
+    logging.debug('opened connection')        
+    logging.debug(f'Going to execute PATCH query on task {taskId} for user {userId}')  
+    #creates a list of fields to update
+    columnsToUpdate = list(task_fields.keys())
+    #params list to later use in sql query
+    params = []
+    fieldsInQuery = ''
+    #iterates through column values and adds them as params to be passed into sql query
+    for column in columnsToUpdate:
+        #accounts for no comma after the last param in the sql guery string
+        if column == columnsToUpdate[-1]:
+            comma = ' '
+        else:
+            comma = ', '
+        #avoids potential sql injection by using ? placeholder for column values
+        fieldsInQuery += "{} = ?{}".format(column, comma)
+        #appends each of columns to the list
+        params.append(task_fields.get(column))
+    #adds userId and taskId to params
+    params.extend([userId, taskId])
+    #set the query body
+    sql_query = """UPDATE [dbo].[tasks] SET {} WHERE userId = ? AND taskId = ?""".format(fieldsInQuery)
+    try: 
+        rowcount = cursor.execute(sql_query, params).rowcount
+        #reports out an unsuccessful result
+        if not rowcount:
+            logging.error('Invalid input and/or no record with the requested parameters exists in db')
+            return func.HttpResponse('Invalid input', status_code=404)                
+        logging.debug('UPDATE query executed')
+        logging.debug(f"Executed the query: {rowcount} rows affected for taskId {taskId}")
+        return func.HttpResponse(status_code=200)
+    except Exception as e:
+        logging.critical('Unable to execute the query')
+        return func.HttpResponse("Error: %s" % str(e), status_code=400)
     finally:
         #commits changes to db
         cnxn.commit()
@@ -80,20 +130,20 @@ def delete(userId, taskId):
     #connects to db
     cnxn = connect()
     cursor = cnxn.cursor()
+    logging.debug('opened connection')        
+    logging.debug(f'Going to execute DELETE query on task {taskId} for user {userId}')
+    sql_query = ("""DELETE FROM [dbo].[tasks] WHERE userId = ? AND taskId = ?""")
     try:
-        logging.debug('opened connection')        
-        logging.debug(f'Going to execute DELETE task {taskId} for user {userId} query')
-        sql_query = ("""DELETE FROM [dbo].[tasks] WHERE userId = ? AND taskId = ?""")
-        try:
-            rowcount = cursor.execute(sql_query, userId, taskId).rowcount
-            if not rowcount:
-                logging.error('No record with the requested parameters exists in db')
-                return func.HttpResponse('No record to delete', status_code=404)                
-            logging.debug(f"Executed the DELETE query: {rowcount} rows affected for taskId {taskId}")
-            return func.HttpResponse(status_code=200)
-        except Exception as e:
-            logging.error('Unable to execute the query')
-            return func.HttpResponse("Error: %s" % str(e), status_code=400)
+        rowcount = cursor.execute(sql_query, userId, taskId).rowcount
+        #returns error in the event of unsuccessful query
+        if not rowcount:
+            logging.error('No record with the requested parameters exists in db')
+            return func.HttpResponse('No record to delete', status_code=404)                
+        logging.debug(f"Executed the DELETE query: {rowcount} rows affected for taskId {taskId}")
+        return func.HttpResponse(status_code=200)
+    except Exception as e:
+        logging.error('Unable to execute the query')
+        return func.HttpResponse("Error: %s" % str(e), status_code=400)
     finally:
         #commits changes to db
         cnxn.commit()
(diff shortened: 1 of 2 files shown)


functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
 #GET API method function
 def get(param1, param2):
     #connects to db
+    logging.debug('Attempting a db conenction')
```
lbrown51, line 13: I feel weird commenting about spelling, but connection

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
-        data = cursor.fetchall()
-        logging.info(f"Got result: {data}")
-        return data
-#irrespective of results, closes connection
+        logging.debug('Executed the query')          
+        row = cursor.fetchone()
+        logging.info(f"Got result: {row}")
+        if not row:
+            logging.error('No record with the requested parameters')
+            return func.HttpResponse('Task not found', status_code=500)
```
lbrown51, line 31: I think this should actually be a 404 status code because the resource wasn't found. It's not really an error on our side per se.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
 #GET API method function
 def get(param1, param2):
```
lbrown51, line 11: Does it make sense to just have the method names be get, update, etc... or should we be doing something like get_user_task, update_user_task, etc (something more descriptive)? Maybe it's obvious given the file that it's in?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
-        # update task: title and description. Client passes userId and taskId, 
-        # for sprint 1 other fields are HARDCODED
-        sql_query = ("""UPDATE [dbo].[tasks]
-        SET title = 'Title updated by team1', description = 'Description updated by team1'
-        WHERE userId = ? AND taskId = ?""")   
-        rowcount = cursor.execute(sql_query, param1, param2).rowcount
-        logging.info(f"Executed the query: {rowcount} rows affected")
-#commits changes to db
+        logging.info('opened connection')        
+        logging.info('Going to execute UPDATE task query')
```
lbrown51, line 48: What's your decision criteria for switching between info and debug?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+        # task_req_body input check
+        logging.debug("Checking the request body for necessary fields to update a task")
+        try:
+            assert 'title' in task_req_body, "New task request body did not contain field: 'title'"
+            assert 'description' in task_req_body, "New task request body did not contain field: 'description'"
+            assert 'dueDate' in task_req_body, "New user request body did not contain field: 'dueDate'"
+        except AssertionError as task_req_body_content_error:
+            logging.critical('New user request body did not contain fields to update a task')
```
lbrown51, line 56: While I agree that db connection issues are more likely than not supposed to be critical, I don't think request errors are. We should probably ask Toddy tonight, but I really think that if it doesn't require changes on our side (i.e. they just need to change the request body), then it's more on the side of an error rather than a critical.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+            return func.HttpResponse(task_req_body_content_error.args[0], status_code=400)
+        logging.info('New user request body contained necessary fields to update a task')  
+        # Unpack task data
+        title = task_req_body.get('title')
+        description = task_req_body.get('description')
+        dueDate = datetime.strptime(task_req_body.get('dueDate'), '%d/%m/%y %H:%M:%S')     
+        # update task: title and description
+        sql_query = """UPDATE [dbo].[tasks]
+        SET title = '{}', description = '{}', dueDate = '{}'
+        WHERE userId = ? AND taskId = ?""".format(title, description, dueDate)
```
lbrown51, line 66: I think instead of format, title, description, etc can also use the ? format. You can also package all your params as a tuple (title, description, dueDate, etc..) and pass them as cursor.execute(sql_query, params).

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
-            #     #createdDate #'20120618 10:34:09 AM' and title/description are hardcoded for sprint1, update with automated date stamp and url params later
-            #     sql_query = ("""INSERT INTO dbo.tasks (userId, title, description, createdDate)
-            #     VALUES (?, ?, ?, '20120618 10:34:09 AM')""")
-            #     cursor.execute(sql_query, userId, 'Do It', 'Almost like Nike motto')   
-            #     logging.info('Executed the query')
-
-#UPDATE API method function
-def update(param1, param2):
+#PUT API method function
+def update(param1, param2, task_req_body):
```
lbrown51, line 42: These params should probably be named after what they are :)

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+        # update task: title and description
+        sql_query = """UPDATE [dbo].[tasks]
+        SET title = '{}', description = '{}', dueDate = '{}'
+        WHERE userId = ? AND taskId = ?""".format(title, description, dueDate)
+        try:
+            rowcount = cursor.execute(sql_query, param1, param2).rowcount
+            if not rowcount:
+                logging.error('No record with the requested parameters exists in db')
+                return func.HttpResponse('No record to update', status_code=500)                
+            logging.debug('UPDATE query executed')
```
lbrown51, line 72: Maybe have this in an else block. I know you don't have to, but it might make reading the code easier?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+            logging.info(f"Executed the query: {rowcount} rows affected for taskId {param2}")
+            return func.HttpResponse(status_code=200)
+        except Exception as e:
```
lbrown51, line 75: What would this exception be caused by?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
 #DELETE API method function
-def delete(param):
-    return ('You selected POST method with {param} values. Functionality under construction')
-            # STARTER CODE FOR NEXT SPRINT, SQL QUERY TESTED
-            #     # #delete row, client passes in userId and taskId
-            #     sql_query = ("""DELETE FROM [dbo].[tasks] WHERE userId = ? AND taskId = ?""")
-            #     cursor.execute(sql_query, userId, taskId)
-            #     logging.info('Executed the query')
-            #     cnxn.commit()
+def delete(param1, param2):
```
lbrown51, line 87: These params should probably be named as what they are :)

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+        # description = task_req_body.get('description')
+        # dueDate = datetime.strptime(task_req_body.get('dueDate'), '%d/%m/%y %H:%M:%S')     
+
+        # #delete row, client passes in userId and taskId
+        sql_query = ("""DELETE FROM [dbo].[tasks] WHERE userId = ? AND taskId = ?""")
+        try:
+            rowcount = cursor.execute(sql_query, param1, param2).rowcount
+            if not rowcount:
+                logging.error('No record with the requested parameters exists in db')
+                return func.HttpResponse('No record to delete', status_code=500)
```
lbrown51, line 115: I think this is probably supposed to be a 404 response, because the resource was not found

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+        sql_query = ("""DELETE FROM [dbo].[tasks] WHERE userId = ? AND taskId = ?""")
+        try:
+            rowcount = cursor.execute(sql_query, param1, param2).rowcount
+            if not rowcount:
+                logging.error('No record with the requested parameters exists in db')
+                return func.HttpResponse('No record to delete', status_code=500)                
+            logging.debug('Executed the DELETE task query')
+            logging.info(f"Executed the query: {rowcount} rows affected for taskId {param2}")
+            return func.HttpResponse(status_code=200)
+        except Exception as e:
```
lbrown51, line 119: What would cause this error?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -81,38 +138,45 @@ def connect():
         db_username = os.environ["ENV_DATABASE_USERNAME"]
         db_password = os.environ["ENV_DATABASE_PASSWORD"]
         driver = '{ODBC Driver 17 for SQL Server}'
+        logging.info('Uses hardcoded ODBC Driver 17 for SQL Server string')
```
lbrown51, line 141: Do you think that this would more lean towards being a debug rather than an info?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -81,38 +138,45 @@ def connect():
...
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            userId = req_body.get('userId')
-            taskId = req_body.get('taskId')
+        logging.debug("Got neither userId nor taskId")
+        userId = req_body.get('userId')
+        taskId = req_body.get('taskId')
```
lbrown51, line 177: What's happening here? What's the reason for looking for the route params in the request body? Is it for recovery / diagnosis if the request was formatted incorrectly?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+    cursor = cnxn.cursor() 
     try:
+        logging.debug('opened connection')        
+        logging.debug('Attempting to execute GET task query')
         #Get task title, description and user name by userId and taskId
-        sql_query = ("""SELECT tasks.title, tasks.description, CONCAT (users.firstName, ' ', users.lastName) AS "user" 
-        FROM [dbo].[tasks] JOIN [dbo].[users] 
-        on [dbo].[tasks].userId = [dbo].[users].userId
-        WHERE [dbo].[users].userId = ? AND [dbo].[tasks].taskId = ?""")
+        sql_query = ("""SELECT tasks.title, tasks.description, tasks.createdDate, tasks.dueDate,
```
farhadbahrehmandhenry, line 20: another shorter alternative =>
"Select * from tasks where userId = ? and taskId = ?"
of course, if we don't want to have the user's name. just a suggestion.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
+            data = dict(zip(columns, row))
+        return func.HttpResponse(json.dumps(data, default=str), status_code=200, mimetype="application/json")
     finally:
         cursor.close()
         cnxn.close()
-        logging.info('Closed the connection')
+        logging.info('Closed the db connection')   
 
-#POST API method function       
-def post(param):
```
farhadbahrehmandhenry, line 31: don't we need to implement POST for this function?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
-        WHERE userId = ? AND taskId = ?""")   
-        rowcount = cursor.execute(sql_query, param1, param2).rowcount
-        logging.info(f"Executed the query: {rowcount} rows affected")
-#commits changes to db
+        logging.info('opened connection')        
+        logging.info('Going to execute UPDATE task query')
+        # task_req_body input check
+        logging.debug("Checking the request body for necessary fields to update a task")
+        try:
+            assert 'title' in task_req_body, "New task request body did not contain field: 'title'"
```
farhadbahrehmandhenry, line 52: why if we need to update only one field? do we have to change the other fields two? what if we want to update the "completed" field?

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: good catch, this is trailing from sprint1

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
```diff
@@ -6,71 +10,124 @@
...
-            #     logging.info('Executed the query')
-            #     cnxn.commit()
+def delete(param1, param2):
+    #connects to db
+    cnxn = connect()
+    cursor = cnxn.cursor()
+    try:
+        logging.info('opened connection')        
+        logging.info('Going to execute DELETE task query')
+        # # task_req_body input check
```
farhadbahrehmandhenry, line 94: you may want to consider deleting these comments. I think these are leftovers from the update function. lines 94 - 107

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
farhadbahrehmandhenry: yes, this part seems unnecessary. I think this part might be leftovers from before.

nataliagourova: Per Microsoft documentation, they are recommending to first look for params in the url, and if not available, in the request body:
[httpTrigger](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python)
[HttpRequest class](https://docs.microsoft.com/en-us/python/api/azure-functions/azure.functions.httprequest?view=azure-python)
With that, the script first looks for route params and if it finds none, it examines the request body.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: I am leaning towards info, as we are aware we're using this driver and it is hardcoded, so we may not want the message pop up each time for when we set the log levels at debug and up

functions/HttpTriggerAPIUsersIdTaskId/function.json
```diff
@@ -8,10 +8,10 @@
...
       "methods": [
         "get",
-        "post",
-        "delete",
-        "update"
-      ]
+        "put",
+        "delete"
+      ],
+      "route": "users/{userId:int?}/task/{taskId:int?}"
```
farhadbahrehmandhenry, line 14: now that I think this route that you got here makes more sense.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Inability to execute the query on line 112

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
farhadbahrehmandhenry: yeah but with this format ("users/{userId:int?}/task/{taskId:int?}") that you got in your function.json, it's almost impossible not to get params, if I'm not mistaken.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: My approach is that as we are not going to have the end user working with this endpoint, the check for the appropriate param should be happening on the client side, which would indicate their software is not working properly / raise 500 error code. If not 500, I'd use 400 "Bad Request" and steer away from 404 in this instance.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: indeed

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Not for this function. The taskId needs to be in place for users/user_id/tasks/task_id in order for the endpoint to be called. it is disabled in the swagger doc from the last quarter, too. ;) Creation should happen in the users/user_id/tasks/ endpoint

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: good call

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Makes sense, and thank you for pointing this out. We could still have a join and select * from tasks, and user name from users. My thinking is along the lines of, if we need to expand the database in the future, we do not have all of the contents of the table exposed through the API, and selecting specific columns will help with that

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: These are all valid questions, and something for which we did not receive specifications. For now my code is built to update all three (title, description, dueDate) or none. I have not considered the other columns for now as they are not populated in the database. This added functionality question may need to be addressed to Toddy / request for clarifications on specs. It could be that the client code will be responsible for getting the fields for the task first, and populating the update request with those fields that do not need to be changed, so that the query can be executed.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Sure, if we draw the parallel with 500 vs. 404 discussion, debug would make sense here vs. critical

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: That's exactly it, my approach was that looking at the parent folder name aka API endpoint makes it clear what those CRUD functions methods are for

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: updated

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: sure, it's all fair game

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Should this be added to the list of questions for Toddy, or do you feel strongly enough to go ahead and delete this section? It is prefaced by "if" and will not execute when we get at least one param, I think it does not hurt to leave it in.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: For level setting, this is Tuesday's section execution: cursor.execute('UPDATE tasks SET **title**=?, **taskDescription**=?, **completed**=? WHERE taskId=?', title, taskDescription, completed, taskID)

nataliagourova: Adding the "completed" field is a quick fix  ;)

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
theKunte: We can clarify this tonight with toddy, but I would agree with Lenny in this case.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
farhadbahrehmandhenry: thanks. this code might be useful in this case to update fields in the body dynamically => 
`    fieldsToUpdate = list(task_req_body.keys())
    updatableFields = ['title', 'description', 'dueDate'] # you can add all fields that apply

    if len(fieldsToUpdate) == 0:
      logging.critical('New user request body did not contain fields to update a task')

      return func.HttpResponse(task_req_body_content_error.args[0], status_code=400)
    else:
      if set(fieldsToUpdate).issubset(updatableFields):
        fieldsInQuery = ''

        for field in fieldsToUpdate:
          comma = ' ' if field ==  fieldsToUpdate[-1] else ', '
          fieldsInQuery += "{} = '{}'{}".format(str(field), str(task_req_body[field]), comma)
 
        sql_query = """UPDATE tasks SET {} WHERE userId = ? AND taskId = ?""".format(fieldsInQuery)
    else:
      # there are non-sense fields in the body we need to handle`

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
farhadbahrehmandhenry: thanks a lot for the explanation. I was thinking that might be a good idea to return a 405 error when we get a request that is not GET, DELETE, or PUT.

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: removed

functions/HttpTriggerAPIUsersIdTaskId/__init__.py
nataliagourova: Updated per Toddy's comments

farhadbahrehmandhenry (approved):

jak-ashuraliev (approved):
Overall, good implementation of comments and logging. Good work, Natalia!

Issue #71, opened 2021-01-22
[BUG] Complete the rest of CRUD  functionality of API Users/{userId}


theKunte:
Update PR#49 

**Is your feature request related to a problem? Please describe.**
The feature for the HTTP trigger Users/UserId was not completed in PR#49 which got merged into the develop branch. The task in this request was to update the Azure function app/ users/userId for all CRUD. ( GET/PUT/POST/PATCH/DELETE) and to add logging,

**Describe the solution you'd like**
I need to implement the functionality for the API

**Describe alternatives you've considered**
No alternatives in this case. These are core functionalities

theKunte:
# Resolves Task/BUG :[#71](https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/71)

# To Test API function users/{userId}
**Test this function on Azure:**
- API call - [users/2](https://nsc-functionsapp-team1.azurewebsites.net/api/users/2)


![image](https://user-images.githubusercontent.com/32443542/106347531-113ef880-6274-11eb-883c-f8338cc3d978.png)



# This code completes the implementation of the Azure function for the previous PR #49  that is already merged into the dev branch

 - The following methods got implemented and Updated: GET/POST/PUT/PATCH/DELETE

- Once this PR gets approved the function will be deployed automatically into the development branch
-  To test this feature locally make sure to have the following: 
      - Azure function extension for VSCode
      - Pull the branch to your machine and make sure to be on the correct branch
      - Install the ODBC driver so you can connect to the database: For any issues with the ODBC driver checkout PR49 for additional documentation to fix ODBC driver
      -In local.settings.json at the root of the functions folder, enter in the values for the keys below in the following format: { "Values": { "KEY": "VALUE" } }
ENV_DATABASE_NAME (name of the database)
ENV_DATABASE_PASSWORD
ENV_DATABASE_SERVER (this is the server location)
ENV_DATABASE_USERNAME
- You can click F5 or click run->Start Debugging in the functions folder to start it up

- The GET request takes the userId and returns all a specific user by their Id. 
- The POST request is not allowed for this call.
- The PUT request takes a JSON object { "firstName": "add your firstname", "lastName": "add your last name", "email": "addYourEmail@email.com" } and will update the user firstName, lastName, email when you select a specific userId. 
- The DELETE request takes the userId and deletes the user from the database.
[...]

PR #49, opened 2021-01-19
Feature/task12 azure function usersId

theKunte:
Task: [#12](https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/12)  

Create and deploy Azure Functions from Visual Studio Code: 

1. Create a serverless app in vscode with extensions and python: 
- Prerequisites: 
    - Create or use existing Azure Subscription
    - Install software to local workstation:
           - Install the Azure Functions Core Tools in Visual Studio Code
           - Install Python and Visual Studio Code with the Azure Function extension. 
           - Sign in to Azure. 
           - Verify your environment: 
           NOTE: Azure function only work with Python version 3.6, 3.7 and 3.8. 
           Reference:  https://docs.microsoft.com/en-us/azure/developer/python/tutorial-vs-code-serverless-python-01
           
           Verify your environment:
           `npm install -g azure-functions-core-tools` 
     
     2. Create a Python function for Azure functions: 
     Reference: https://docs.microsoft.com/en-us/azure/developer/python/tutorial-vs-code-serverless-python-02
     Note: When creating a HttpTriger in Visual Studio Code open the function folder in a separate window in Visual Studio Code and create your function in there. This will help to structure your Triggers and keep files in the same folder instead of in the root folder
     
 **3. Examine the Python code files in vscode:**
  Reference:

-  https://docs.microsoft.com/en-us/azure/developer/python/tutorial-vs-code-serverless-python-03

  ### 4. Debug the Azure Functions Python Code locally:
  **Reference:** 

- https://docs.microsoft.com/en-us/azure/developer/python/tutorial-vs-code-serverless-python-04?tabs=powershell

### 5. Deploy Azure Function in Python
 Reference: https://docs.microsoft.com/en-us/azure/developer/python/tutorial-vs-code-serverless-python-05
  
**Purpose:** This feature implemented the API endpoint for getting a specific user by userId. 

**TEST PULL REQUEST:**
### **Requirements to set up Azure Function and Deploy them to Azure:** 
  - Clone repository: 
  -  - Install Azure Functions extension in vscode.
      - Open the functions folder separately in vscode.
              Note: if you have the root repository folder open the _azure.function_ extension will not work.
     
      -  Install **Python version 3.8** if you haven't yet 
      -  Note: Azure function does not work with Python 3.9  See documentation: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-python
      -  Set up environment in vscode/ .venv
      -  Sign in to Azure 
[...]

PR #89, opened 2021-01-29
Feature/team1 issue71 complete azure users user id function

theKunte:
**Resolves Task/BUG** :[#71](https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/71)

**To Test API function users/{userId}** 
- Test this function on Azure:
- API call - [users/1](https://nsc-functionsapp-team1.azurewebsites.net/api/users/1)

- This code completes the implementation of the Azure function for the previous PR #49  that is already merged into the dev branch
           - The following methods got implemented and Updated: GET/POST/PUT/PATCH/DELETE

- Once this PR gets approved the function will be deployed automatically into the development branch
-  To test this feature locally make sure to have the following: 
      - Azure function extension for VSCode
      - Pull the branch to your machine and make sure to be on the correct branch
      - Install the ODBC driver so you can connect to the database: For any issues with the ODBC driver checkout PR49 for additional documentation to fix ODBC driver
      -In local.settings.json at the root of the functions folder, enter in the values for the keys below in the following format: { "Values": { "KEY": "VALUE" } }
ENV_DATABASE_NAME (name of the database)
ENV_DATABASE_PASSWORD
ENV_DATABASE_SERVER (this is the server location)
ENV_DATABASE_USERNAME
- You can click F5 or click run->Start Debugging in the functions folder to start it up

- The GET request takes the userId and returns all a specific user by their Id. 
- The POST request is not allowed for this call.
- The PUT request takes a JSON object { "firstName": "add your firstname", "lastName": "add your last name", "email": "addYourEmail@email.com" } and will update the user firstName, lastName, email when you select a specific userId. 
- The DELETE request takes the userId and deletes the user from the database.
-To test using JSON, copy the link provided by the debugger (the "/users/{userId}" ) into Postman, and use as usual for GET and PUT (info goes into the body), DELETE (userId). PUT and PATCH are not allowed
- Error handling and Debug handling included in this feature

DATE | ACTIVITY | TIME
-- | -- | --
1/21 | Review PR49 after class to get an idea of what I need to update | 60 minutes
1/23 | More research of pyodbc Crud functionality | 90 minutes
1/24 | encounter issue with running function locally | 120 minutes
1/24 | Still figuring out what was going on with the issue| 60 minutes
1/24 | Started working on functionality| 60 minutes
1/25 | Setup all rest methods (get, post, put, delete) to work as expected | 120 minutes
1/26 | Added error handling to post requests | 90 minutes
1/26 | Added storage account to function app on azure | 20 min
1/28 | Setup all rest methods (get, post, put, delete, patch) to work as expected | 230 minutes
1/28 | Added more error handling for the DB connection logging | 60 minutes
1/28 | wrote this PR which I will have to update again |60 minutes


functions/HttpTriggerAPIUsersId/function.json
```diff
@@ -8,7 +8,9 @@
       "name": "req",
       "methods": [
         "get",
-        "post"
+        "post",
+        "put",
```
jak-ashuraliev, line 12: POST and PUT methods are defined - Looks good!

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -11,6 +11,15 @@
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    logging.info(
+        'Python HTTP trigger for users/userId is processing a request ')
+
+    # Check request method
+    if not req.method:
+        logging.critical('No method available')
+        raise Exception('No method passed')
```
lbrown51, line 20: This should probably return some http response telling the user that they didn't send a method. Maybe 405 or 400.

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -23,30 +32,48 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
...
+                elif req.method == 'PATCH':
+                    return patch()
+                elif req.method == 'DELETE':
+                    return delete(req, conn, id)
 
+                conn.close()
+                logging.debug('Connection to DB closed')
 
-def get_user_by_id(conn, id):
+        except Exception as e:
```
lbrown51, line 61: This blanket catch for exceptions should basically work, but doesn't really give us a lot of (in-code) information about what kinds of errors could arise. Comments would about this would be nice but you could also do a separate except for known errors?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+    return func.HttpResponse(
+        "Method not allowed!",
+        status_code=405
+    )
+
+
+def put(req, conn, id):
+    user_req_body = req.get_json()
+    logging.debug("Verifying fields in request body to update a user by ID")
+    try:
```
lbrown51, line 107: Have we decided that they have to have all the fields in their request even if they are only updating one of  them?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+
+def post():
+    logging.debug("POST is not implemented")
+    return func.HttpResponse(
+        "Method not allowed!",
+        status_code=405
+    )
+
+
+def put(req, conn, id):
```
lbrown51, line 104: Maybe the param should be user_id rather than id. I know that it's obvious in this file that we're only really using user ids but in general two different kinds of ids show up in the functions so there might still be some confusion.

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+    logging.debug("User request body contains all the necessary fields!")
+    with conn.cursor() as cursor:
+        # Unpack user data
+        firstName = user_req_body["firstName"]
+        lastName = user_req_body["lastName"]
+        email = user_req_body["email"]
+        user_params = (firstName, lastName, email)
+
+        # Update user  query
+        update_user_query = "UPDATE dbo.users SET firstName = ?, lastName = ?, email = ? WHERE userId={}".format(
```
lbrown51, line 125: What's the purpose for using the format instead of just another ?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+                    status_code=404
+                )
+            logging.debug("Executing query: " + update_user_query)
+            cursor.execute(update_user_query, user_params)
+            logging.debug("User was updated successfully!.")
+            return func.HttpResponse(
+                "User updated",
+                status_code=200
+            )
+        except Exception as e:
```
lbrown51, line 145: It would be nice if this was a more specific exception rather than just Exception. Though, to be fair, I think I also used a generic exception for this failing point as well.

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+        status_code=405
+    )
+
+
+def delete(req, conn, id):
+    logging.debug("Attempting to retrieve user by ID and delete the user...")
+    with conn.cursor() as cursor:
+
+        # Delete user  query
+        delete_user_query = "DELETE FROM dbo.users  WHERE userId={}".format(
```
lbrown51, line 166: Why use format rather than "?"?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+            id)
+
+        try:
+            logging.debug("Check if userId exists in database: " + id)
+            cursor.execute(
```
lbrown51, line 171: Maybe you should write a complete sql function that does this checking and deleting if it exists functionality? That might be too much work for little reward?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+                    status_code=404
+                )
+            logging.debug("Executing query: " + delete_user_query)
+            cursor.execute(delete_user_query)
+            logging.debug("User was deleted successfully!.")
+            return func.HttpResponse(
+                "User deleted",
+                status_code=200
+            )
+        except Exception as e:
```
lbrown51, line 186: Again same thing with the generic exception :)

lbrown51 (commented):
A general comment is that we probably don't want any of the react code in this PR. Maybe you just need to update from upstream/development?

jak-ashuraliev (approved):
Since I cannot actually run the functions on my local machine, I can only make comments. Overall, good logging implementation!

functions/HttpTriggerAPIUsersId/__init__.py
farhadbahrehmandhenry: update_user_query = "UPDATE dbo.users SET firstName = ?, lastName = ?, email = ? WHERE userId=?"

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+            "User request body did not contain the necessary fields!")
+        return func.HttpResponse(user_req_body_content_error.args[0], status_code=400)
+
+    logging.debug("User request body contains all the necessary fields!")
+    with conn.cursor() as cursor:
+        # Unpack user data
+        firstName = user_req_body["firstName"]
+        lastName = user_req_body["lastName"]
+        email = user_req_body["email"]
+        user_params = (firstName, lastName, email)
```
farhadbahrehmandhenry, line 122: user_params = (firstName, lastName, email, id)

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+
+        # Update user  query
+        update_user_query = "UPDATE dbo.users SET firstName = ?, lastName = ?, email = ? WHERE userId={}".format(
+            id)
+
+        try:
+            logging.debug("Check if userId exists: " + id)
+            cursor.execute(
```
farhadbahrehmandhenry, line 130: here you are making sure that the user exists in the database then proceed to update? in this case, it shouldn't be necessary, because you'll get a 404 automatically if you want to update a user that doesn't exist.

functions/HttpTriggerAPIUsersId/__init__.py
farhadbahrehmandhenry: delete_user_query = "DELETE FROM dbo.users  WHERE userId=?"

functions/HttpTriggerAPIUsersId/__init__.py
farhadbahrehmandhenry: not sure this is necessary, you probably will get a 404 if you want to delete a user that doesn't exist.

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -58,8 +85,107 @@ def get_user_by_id(conn, id):
...
+                "SELECT * FROM users WHERE userId={}".format(id))
+            row = cursor.fetchone()
+            if not row:
+                return func.HttpResponse(
+                    "User not found",
+                    status_code=404
+                )
+            logging.debug("Executing query: " + delete_user_query)
+            cursor.execute(delete_user_query)
```
farhadbahrehmandhenry, line 180: cursor.execute(delete_user_query, id)

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: That is a good point, but I was focusing on catching all errors regardless of pyodbc or anything else. I could implement better error handling. But in this part, any error that occurs should be a 500 error.

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: This would be a patch request if you want to update something specific. in this case for Put I want to update all fields

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I updated it to user_id!

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: Its just how I had it and it works :D 
So I think this is okay the way it is

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I used this in Sprint 1 when I created the get request. Is there something I am missing why would we use ? ? rather than a string format? The reason why I used string format was that it is an easy way to insert variables into a string. If you have a concern please share it

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I thought about doing that and safe on lines of code... but didn't end up doing it. I might update it :D 
Thanks for pointing that out!

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -11,6 +11,15 @@
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    logging.info(
+        'Python HTTP trigger for users/userId is processing a request ')
+
+    # Check request method
+    if not req.method:
```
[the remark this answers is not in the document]
theKunte: It will be removed in the next commit

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -23,30 +32,48 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
     id = req.route_params.get('userId')
     if not id:
```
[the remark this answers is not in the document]
theKunte: Yup! You are right. I removed this part since it is not necessary

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -23,30 +32,48 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
...
+
+                if req.method == 'GET':
+                    return get(conn, id)
+                elif req.method == 'POST':
+                    return post()
+                elif req.method == 'PUT':
+                    return put(req, conn, id)
+                elif req.method == 'PATCH':
+                    return patch()
+                elif req.method == 'DELETE':
```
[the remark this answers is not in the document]
theKunte: That's a good point. What do you think of a switch statement?

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -23,30 +32,48 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
...
+            return func.HttpResponse(
+                "Internal Server Error",
+                status_code=500
+            )
+
+
+def get(conn, id):
+    logging.debug("Attempting to retrieve user by ID...")
     with conn.cursor() as cursor:
         cursor.execute(
```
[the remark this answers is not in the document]
theKunte: Are you trying to say not to use the string format ?

functions/HttpTriggerAPIUsersId/__init__.py
farhadbahrehmandhenry: that's totally up to you, but the suggested format is to avoid SQL injection.

functions/HttpTriggerAPIUsersId/__init__.py
farhadbahrehmandhenry: to avoid SQL injection.

functions/HttpTriggerAPIUsersId/__init__.py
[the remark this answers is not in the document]
farhadbahrehmandhenry: switch would work too.

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: Thanks Farhad :D My memory is pretty bad nowadays haha!

functions/HttpTriggerAPIUsersId/function.json
nataliagourova: Are you  planning to implement the PATCH in addition to PUT per Toddy's comments?

functions/HttpTriggerAPIUsersId/__init__.py
nataliagourova: Per Toddy's comment, this is not needed

functions/HttpTriggerAPIUsersId/function.json
theKunte: I will add this to the functions.json file but patch wont be allowed since we follow the lenny's swagger doc

farhadbahrehmandhenry (commented):
I just noticed merging you branch will cause deleting the react app from NSC repo, we need to update you branch with NSC first.

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I have updated this!

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I updated the code to avoid SQL injections. Thanks !

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: I have not included any specific error handling. This can be updated later if needed

functions/HttpTriggerAPIUsersId/__init__.py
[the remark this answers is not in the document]
theKunte: I was not able to use switch but I made some adjustment so let me know what you think now.

functions/HttpTriggerAPIUsersId/__init__.py
theKunte: updated

farhadbahrehmandhenry (approved):

farhadbahrehmandhenry:
tested, all the functionalities are working, good to merge.

functions/HttpTriggerAPIUsersId/__init__.py
nataliagourova: Jenni, just an FYI here, if the client wants to update firstName only here via patch, and not supply lastName and email, they will get an error/so one could argue this is not a true patch

PR #109, opened 2021-02-01
Automate Alert Rule Creation

unicornsleuth:
To view existing Alert Rules:
`Get-AzMetricAlertRuleV2 -ResourceGroupName nsc-functions-team1`

**Overview**
This PR creates a script that will create a new Alert Rule to check for 5xx response codes every 5 minutes, and send an alert if the count exceeds 30.
The script is set up so that only the first 5 parameters are required, and it will default to creating the described Rule, but any of the parameters can be added into the original script call to overwrite defaults.
This PR also edits parameter names in create_vnet.ps1 to conform to agreed-upon standards.

**To Test**
Navigate to /Automation/alert
`.\create_alert.ps1 -TenantId <tenant id> -SubscriptionId <subscription id> -SPApplicationId <service principal app id> -SPSecret <service principal secret> -AlertName <alert name>`

*Delete Resource*
Go to the Azure Portal and search for Function App - select `nsc-functionsapp-team1`

image

Scroll to the Alert blade on the left, then click on "Manage alert rules"
Observe that (unless you used custom settings) the conditions for your new rule are the same as the existing http-5xx-alert
Delete your newly created rule to save on resources : )


zxkevinxz (approved):
Works well, makes the proper alert rule.  Was able to compare to the other rule that was there.

Automation/alert/create_alert.ps1
```diff
@@ -0,0 +1,62 @@
...
+} elseif (!$actionResourceGroupExists) {
+    Write-Host "Resource Group $ActionResourceGroupName does not exist. Cannot create Alert."
+} else {
+    # Creates Alert if one of the same name does not already exist in the Resource Group
+    $alertExists = (Get-AzMetricAlertRuleV2 -Name $AlertName -ResourceGroupName $TargetResourceGroupName -ErrorAction SilentlyContinue).Name -eq $AlertName
+    if (!$alertExists) { 
+        Write-Host "Alert does not exist. Creating now."
+        # create Alert with given name
+        New-AzResourceGroupDeployment `
+            -ResourceGroupName $TargetResourceGroupName `
```
jleeboyd, line 44: Could use splatting to pass these parameters later on to make it look nicer, but not a big deal if not.

jleeboyd (approved):
Running the script successfully created a new alert in Azure!

Issue #20, opened 2021-01-11
[Task 6] – Implement script to create Application Gateway


glanzmichael25:
Implement script to create Application Gateway

PR #8, opened 2021-01-09
Task7 azurestoragescript

theanticrumpet:
Resolves https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/7

PR #118, opened 2021-02-03
AppGateway

glanzmichael25:
Includes AppGateway ps1
Includes template json

### DIFF
--- a/Automation/Task6/AppGateway.ps1
+++ b/Automation/Task6/AppGateway.ps1
@@ -1,0 +1,64 @@
+#Creates Azure Application Gateway 
+
+param (
+    [Parameter(Mandatory=$true)]
+    [string]
+    $applicationGatewayName,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $location,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $backendAddressPools,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $backendHttpSettingsCollection,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $frontendIPConfigurations,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $gatewayIPConfigurations,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $frontendPorts
+    
+    [Parameter(Mandatory=$true)]
+    [string]
+    $httpListeners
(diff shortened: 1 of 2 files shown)


Automation/Task6/AppGateway.ps1
```diff
@@ -0,0 +1,64 @@
...
+    [string]
+    $frontendIPConfigurations,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $gatewayIPConfigurations,
+
+    [Parameter(Mandatory=$true)]
+    [string]
+    $frontendPorts
```
ahmadafrasiyab, line 30: There needs to be comma separation between parameters. This is causing issues while running the script.

ahmadafrasiyab:
It would be very helpful if you give some basic steps to test your script including the script command that you are using. Also, when you ask for IP pools, documentation would again be needed to see what format is the input expected in. I would start with writing a basic step by step process to test your script. Including but not limited to what is expected as input to specific parameters that are not very common as opposed to location names or resource group names where it is fairly straightforward to guess the inputs.

Automation/Task6/AppGateway.ps1
```diff
@@ -0,0 +1,64 @@
...
+    
+    [Parameter(Mandatory=$true)]
+    [string]
+    $httpListeners
```
theanticrumpet, line 34: And here

Automation/Task6/AppGateway.ps1
```diff
@@ -0,0 +1,64 @@
...
+    
+    [Parameter(Mandatory=$true)]
+    [string]
+    $requestRoutingRules
```
theanticrumpet, line 38: And here

theanticrumpet:
Did you try running your script at all? The [linked issue](https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/20) has no time tracking so I'm curious how much time was spent actually testing this. There's also no logic for logging into Azure in the script itself which is something [I did wrong the previous sprint](https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8#discussion_r557724464). The point of the automation stuff seems to be scripts that can just run on any given machine with minimal setup.

This could just be on me and not an actually expected thing but I personally don't know what a lot of the parameters the script is asking for off the top of my head and wouldn't be able to test those given no instructions or default testing values.

PR #120, opened 2021-02-04
Issue 94 React user task page

AllisonBC:
**Overview**
This PR creates a page showing a single specified task for user.

**To Test**
http://localhost:3000/users/3/tasks/2


zxkevinxz (approved):
This works as intended. If this was an actual web development class, I would say try not to use var, while it is still being used, I believe best practices prefer using lets and const. Also, to add some kind of handling for when an incorrect endpoint is put in like if they put /users/abc123/tasks/1s2. I noticed no one put in any handling for that, and im not sure if it is even in the scope of this class.

ui/react-ui/src/components/routes/userSingleTask/UserSingleTask.js
```diff
@@ -0,0 +1,40 @@
...
+    fetch(`https://nsc-functionsapp-team1.azurewebsites.net/api/users/${userId}/tasks/${taskId}`)
+    .then(response => response.json())
+    .then(data => setUserSingleTask(data))
+    .catch((error) => console.error(error))
+  }, [userId, taskId]);
+
+  return (
+    <Container>
+      <h3>Selected task</h3>
+      <Table striped bordered hover className="text-left">
```
farhadbahrehmandhenry, line 22: here you could use BootstrapTable component to render the table, it's totally up to you though. if you use that component, it'll have these benefits=>
- you'll avoid hardcoded values
- if in the future some fields get added/removed from the tables, you always get the latest fields.

ui/react-ui/src/components/routes/home/Home.js
```diff
@@ -95,7 +102,7 @@ class Home extends React.Component {
                       return (
                         <input 
                           className='param-input' 
-                          id={endpoint.key} type='text' 
+                          id={`${endpoint.key}-${param.key}`} type='text'
```
farhadbahrehmandhenry, line 105: thanks for updating this line. also as I mentioned in the slack channel, if you replace the line 62 with this 
`if (param.value) updatedEndpoint = updatedEndpoint.replace(param.regex, param.value);`
the bug that we have in there to add the params in the endpoint will be fixed

ui/react-ui/src/components/routes/userSingleTask/UserSingleTask.js
AllisonBC: I thought my implementation clashed with how the BootstrapTable component was set up. We could chat more about it if you want

ui/react-ui/src/components/routes/home/Home.js
AllisonBC: Should be fixed.

ui/react-ui/src/components/routes/userSingleTask/UserSingleTask.js
farhadbahrehmandhenry: don't worry about it, it's already working properly. I'm gonna go ahead and approve you pr.

farhadbahrehmandhenry (approved):

Issue #34, opened 2021-01-14
[BUG] Issues in Pull Request #8


theanticrumpet:
See: https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8

1. Hard Coded names for ResourceGroup, Location, and StorageAccountName:
https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8#discussion_r557723262

2. Missing in-script Azure sign-in:
https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8#discussion_r557724464

3. Using ARM template ~~to check if a resource already exists~~:
https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8#discussion_r557725413

4. Fix script to work with CI/CD and other automation systems:
https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/8#discussion_r557726074

theanticrumpet:
Regarding point 1: PowerShell `param( ... )` defines variables and has built-in functionality for both accepting command-line arguments and prompting for it.

theanticrumpet:
Regarding point 3: Not "Check if resource already exists" but to use the ARM template to create the resource in the first place.

theanticrumpet:
In sprint 2.5 meeting using [Azure Modules](https://docs.microsoft.com/en-us/powershell/scripting/developer/module/how-to-write-a-powershell-script-module?view=powershell-7.1) for repeated code was mentioned as best practice instead of using `../moreCode.ps1`

Look into this?

theanticrumpet:
Required:
* Powershell
* PS Module Az version 0.7 or later: `Get-InstalledModule -Name "Az"` to check
    *  See https://docs.microsoft.com/en-us/powershell/azure/install-Az-ps?view=azps-5.3.0
* Tenant ID, Application ID, Secret, and Subscription ID found in the Azure console and key vault
-----
Script can be found in [REPO]\Automation\StorageAccount\create_Azure_StorageAccount.ps1
Template can be found in [REPO]\Automation\StorageAccount\template.json

Open up a Powershell Window
Navigate to the location of the script
Type `& create_Azure_StorageAccount.ps1`
Optionally also include parameters: `-tenantId ... -applicationId ... -secret ... -subscriptionId ... -resourceGroup ... -location ... -storageAccountName ...`
Run the command
[...]

PR #113, opened 2021-02-02
Create Login Module

unicornsleuth:
Converts the login.ps1 file to a module

*To Implement*
Find the Login directory under /Automation
Add to your script the line `Import Module ..\Login` - may need to edit if your script is stored differently than the other resource creation scripts
To log in, add the following line to your script:
`Login $TenantId $SPApplicationId $SPSecret $SubscriptionId` (or if your scripts variables are different, use those)

PR #121, opened 2021-02-04
Task34 create storage account script

theanticrumpet:
Resolves: Issue https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/34

* Parameters no longer hard coded, now takes slew of parameters using PowerShell `param( ... )` block at start of file.
* Added functionality from Pull Request https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/pull/113 to sign in to Azure
* Now creates storage account from ARM template
* Removed use of `Read-Host`


theanticrumpet:
> 
> 
> Required:
> 
>     * Powershell
> 
>     * PS Module Az version 0.7 or later: `Get-InstalledModule -Name "Az"` to check
>       
>       * See https://docs.microsoft.com/en-us/powershell/azure/install-Az-ps?view=azps-5.3.0
> 
>     * Tenant ID, Application ID, Secret, and Subscription ID found in the Azure console and key vault
> 
> 
> Script can be found in [REPO]\Automation\StorageAccount\create_Azure_StorageAccount.ps1
> Template can be found in [REPO]\Automation\StorageAccount\template.json
> 
> Open up a Powershell Window
> Navigate to the location of the script
> Type `& create_Azure_StorageAccount.ps1`
> Optionally also include parameters: `-tenantId ... -applicationId ... -secret ... -subscriptionId ... -resourceGroup ... -location ... -storageAccountName ...`
> Run the command
> Script will check if the resource group exists, create it if it doesn't, and make sure the location is set properly
> Script will check if the given storage account name [is valid](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage)
> 
>     * If the name is valid script will create the storage account using the given name, resource group, and location
> 
>     * If not valid script will output the message contents of `Get-AzStorageAccountNameAvailablility -Name $name` (usually either already exists or didn't use exclusively lowercase letters)
>       Output of script is saved in a text file `create_Azure_StorageAccount.log`
>       Before running script again to test recommend opening [the Azure Console](https://portal.azure.com/#home) and deleting either the Resource Group or the Storage Account you just made
> 
> 
> Possible changes:
> 
>     * Dynamically building resource group name? Still not really a fan of this but if we're going forward this way easy add
> 
>     * Further template lines for all the nitty-gritty Storage Account details? Current `template.json` seems to work just fine but there's _a lot_ of options to change in storage accounts

ahmadafrasiyab:
I was able to run your script and successfully create a storage account. No errors. Results:
image

Automation/StorageAccount/create_Azure_StorageAccount.ps1
```diff
@@ -0,0 +1,66 @@
...
+)
+
+Start-Transcript -Path "$PSScriptRoot\create_Azure_StorageAccountlog.log"
+Write-Host "Logging to $PSScriptRoot\create_Azure_StorageAccountlog.log"
+Write-Host "Create Azure Storage Account"
+
+# Custom module Login
+Import-Module ..\Login
+# Provides function Login (tenantID, applicationID, secret, subscriptionID)
+Login $tenantId $applicationId $secret $subscriptionId
```
ahmadafrasiyab, line 28: I used the Connect-AzAccount command, didnt know there was this Login module too. Works great.

ahmadafrasiyab (approved):
Scripts runs and produces storage account as expected.

Automation/StorageAccount/create_Azure_StorageAccount.ps1
theanticrumpet: Yeah, this is brand new! It should be a very quick fix to add it to all the scripts. Especially since it is just some logic that takes those four keys and runs the `Connect-AzAccount` cmdlet using service principals

jinudean: good to know Logi-in module

jinudean:
good to know login

jinudean (approved):

PR #123, opened 2021-02-04
[BUG] Feature HttpTriggerAPIUsers/{UserId} DB error

theKunte:
# [BUG] DB error with HttpTrigerAPIUsers/{UserId}

I encountered an issue with the DB. 
After deployment, the HttpTrigger for Users/{userId} was working temporarily. However, after some time it stopped working. I am not sure if this will fix this issue. 
I removed the pypyodbc dependency from the requirements file and from the HttpTrigger for Users/{userId}


Kemet2016 (approved):
I think that made it work better than pypyodbc

farhadbahrehmandhenry (approved):

PR #124, opened 2021-02-04
[ Task ] Serverless-artillery report for /users/{user_id}/tasks

jaguiniga:
Server less Artillery test and report. For Task Id


StopMakingUsernamesUnique (approved):
Looks good, but I would recommend adding create function when you can. I also think you should remove stage name tags, and add region tags on all sls commands, but i may be mistaken.  Ill need to talk to kevin, but I think the --region tags will work with just the $AWS_SLSART_REGION var since we have defined the aws-region above.

chonger878:
Were we supposed to have like some kind of result.json file or something like that?  Otherwise, everything else looks great and straightforward.

chonger878 (approved):
This was easy to follow and I feel like I still got a better idea of how the performance test works as I look at yours.

PR #125, opened 2021-02-04
Team 4 task: Serverless -artillery report for endpoint \users

chonger878:
Add a github action that will run a serverless-artillery report for the endpoint user.  This can be used to generate a list of users that can be pulled into a json file or data bucket.


StopMakingUsernamesUnique (approved):
Great, try including a -rampTo phase so we can stress test

StopMakingUsernamesUnique (approved):
I don't think we are using the stage names anymore, I think we are using region instead, check the perf test kevin has that is now working

chonger878:
@StopMakingUsernamesUnique , I can't seem to locate that perf test.  What issue did it cover?

jaguiniga (approved):
Other than the permissions it looks like it should work.

PR #167, opened 2021-02-17
Create deletion script for S3 buckets - task154

### LINKED ISSUE #154: [Task] Create a script to delete obsolete buckets from S3
While there exists a script to delete obsolete Azure resources, no such script exist for deleting obsolete S3 buckets (which we have a lot of). This should remedy that problem.

Please create a script that lets the script runner delete obsolete buckets (and gives them the necessary information in order to make that decision)

Delete them one by one on the S3 console.

lbrown51:
Resolves #154 

- This code adds a script for deleting obsolete buckets on S3
- Features
  - Lists all the buckets
  - Either lists the bucket owner's email or indicates that it doesn't have an owner
  - Either lists the objects in the bucket or indicates that it doesn't have objects
  - Gives the user the option to exit the script at any point and still proceed to delete previously scheduled buckets (you don't have to do it all at once)
- It doesn't automatically delete empty buckets or buckets that don't have tags (should it?)
- Must be run locally
- To run
  - Install required packages 
    - Navigate to Automation/S3 and use `pip install -r ./requirements.txt`
    - Also, you could just type `pip install boto3 colorama` 
  - From Automation/S3
    - Enter `python .\delete_s3_buckets.py --AccessKeyId access_key --SecretAccessKey secret`
    - You can set a custom region with `--RegionName region_name`
    - Use `python .\delete_s3_buckets.py -h` to get a super simple help message
    - You can find your Access Key ID and Secret Access Key in the message from Toddy last quarter. You can at least get the access key in the IAM management console on AWS (click: IAM -> Users -> Your email -> Security credentials), but won't be able to get the secret
- All the errors I'm aware of are taken care of using exceptions so there shouldn't be any troubleshooting necessary.


theKunte (approved):
I liked the option for exiting the script! Nice work. 
Tested it with a new bucket and works as described. Approved on my end!

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
+import boto3
+from botocore.config import Config
+from botocore.exceptions import ClientError, NoCredentialsError
+from colorama import init, Fore
+import argparse
+
+COL_SIZE = 15
+init(autoreset=True)
+
+parser = argparse.ArgumentParser(description="Deletes S3 buckets.")
```
toddysm, line 7: In your PR comment you wrote that it doesn't require parameters - what did you mean with that?

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+
+        bucket_owner_email = "No Owner"
+        if len(bucket_owner_email_list) == 1:
+            bucket_owner_email = bucket_owner_email_list[0]['Value']
+
+        return bucket_owner_email
+
+    except s3_client.exceptions.from_code('NoSuchTagSetError') as e:
+        print(Fore.YELLOW + "Bucket does not have any tags!")
```
toddysm, line 133: I would still return some value else you will end up with a lot of `null` strings on the console/logs.

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+            print(Fore.GREEN + f'{"OwnerEmail:":<{COL_SIZE}}{bucket_owner_email}')
+            bucket_object_list_prepared_for_deletion = []
+            if len(bucket_object_list) != 0:
+                print(Fore.GREEN + "Bucket Objects:")
+                for bucket_object in bucket_object_list:
+                    # Prepare just in case we want to delete the bucket
+                    bucket_object_list_prepared_for_deletion.append({
+                        'Key': bucket_object
+                    })
+                    print(bucket_object)
```
toddysm, line 69: This is an odd `print` statement. It will be helpful to add some context. What is the purpose of it?

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+            
+            # Prompt user for response
+            delete_bucket_response = ""
+            while delete_bucket_response == "":
+                deletion_prompt_response = input(Fore.YELLOW + "Delete bucket: (y/N/exit)")
+                if deletion_prompt_response == "y":
+                    delete_bucket_response = "True"
```
toddysm: Hmmm... why not just keep the `y`? What is the purpose of reassigning to a string value of `True`?

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+                elif deletion_prompt_response == "N":
+                    delete_bucket_response = "False"
```
toddysm: Same as above...

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+                elif deletion_prompt_response == "exit":
+                    delete_bucket_response = "Exit"
```
toddysm: Same as above...

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+                deletion_information = {
+                    'BucketName': bucket_name,
+                    'Objects': bucket_object_list_prepared_for_deletion
+                }
+                buckets_to_delete.append(deletion_information)
+                print(Fore.GREEN + "Object scheduled for deletion!")
+            elif delete_bucket_response == "Exit":
+                break
+
+            print()
```
toddysm: What is the purpose of this `print` statement?

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+
+            
+            if delete_bucket_response == "True":
+                deletion_information = {
+                    'BucketName': bucket_name,
+                    'Objects': bucket_object_list_prepared_for_deletion
+                }
+                buckets_to_delete.append(deletion_information)
+                print(Fore.GREEN + "Object scheduled for deletion!")
+            elif delete_bucket_response == "Exit":
```
toddysm: What if I type `blah`?

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+init(autoreset=True)
+
+parser = argparse.ArgumentParser(description="Deletes S3 buckets.")
+parser.add_argument('--AccessKeyId', dest='aws_access_key_id', required=True)
+parser.add_argument('--SecretAccessKey', dest='aws_secret_access_key', required=True)
+parser.add_argument('--RegionName', dest='region_name', default='us-west-2', help='Region Name (default: us-west-2)')
+
+args = parser.parse_args()
+
+def main(aws_access_key_id, aws_secret_access_key, region_name):
```
toddysm, line 14: Would be nice to have a doc comment documenting the function including inputs and outputs.

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+                    'Objects': bucket_object_list_prepared_for_deletion
+                }
+                buckets_to_delete.append(deletion_information)
+                print(Fore.GREEN + "Object scheduled for deletion!")
+            elif delete_bucket_response == "Exit":
+                break
+
+            print()
+        
+        print()
```
toddysm: Same here...

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+    except (ClientError, NoCredentialsError)  as e:
+        print(Fore.RED + f'Error connecting to S3! {e}')
+    
+
+def get_owner_email(s3_client, bucket_name):
+    '''
+    Gets the owner's email of an S3 bucket if that information is included in the bucket tags as OwnerEmail
+
+        Parameters:
+            s3_client (S3Client): the client necessary to execute S3 commands
```
toddysm: Take a look at the following guide: https://realpython.com/documenting-python-code/ it has good suggestions for comments and documenting your code

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+
+        return bucket_object_list
+
+    except s3_client.exceptions.from_code('NoSuchBucket') as e:
+        print(Fore.RED + "Bucket does not exist!")
+    
+
+def delete_buckets(s3_client, buckets_to_delete_list):
+    '''
+    Deletes the S3 buckets and their objects
```
toddysm: :) this is not very helpful comment - it is obvious from the name of the function. Maybe adding some details about the steps taken to delete the bucket will be helpful.

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+                        'Objects': bucket_to_delete['Objects']
+                    }
+                )
+                print(Fore.GREEN + "Objects deleted!")
+            
+            bucket_deletion_response = s3_client.delete_bucket(
+                Bucket=bucket_to_delete['BucketName']
+            )
+
+            print(Fore.GREEN + "Bucket deleted! \n")
```
toddysm: Would be nice to add information what was the name of the bucket deleted. The printed message will be much more helpful.

Automation/S3/delete_s3_buckets.py
```diff
@@ -0,0 +1,194 @@
...
+
+    except Exception as e:
+        print(Fore.RED + f'Error while deleting buckets:  {e}')
+
+main(args.aws_access_key_id, args.aws_secret_access_key, args.region_name)
```
toddysm: Are you sure this is correct? This seems strange to me. Check this one: https://realpython.com/python-main-function/

Automation/S3/delete_s3_buckets.py
lbrown51: This is my bad. When I originally wrote the PR, I was using the AWS CLI to configure these parameters and forgot to change the line that says  it doesn't need any when I switched to argparse.

Automation/S3/delete_s3_buckets.py
lbrown51: This is so that the user can check out what objects are in the bucket before choosing to delete. I can comment it, but is that useful as far as making that decision goes?

Automation/S3/delete_s3_buckets.py
lbrown51: Good point, I'll switch it over!

Automation/S3/delete_s3_buckets.py
lbrown51: Good point, I'll switch it over!

Automation/S3/delete_s3_buckets.py
lbrown51: Good point, I'll switch it over!

Automation/S3/delete_s3_buckets.py
lbrown51: There's an input loop above, it will keep asking until you enter one of the defined options. I should print something out if they get it wrong!

Automation/S3/delete_s3_buckets.py
lbrown51: It was to make a newline, but I'll just add that at the start of each loop

Automation/S3/delete_s3_buckets.py
lbrown51: I put it in a if __name__ == "__main__": block, which I think is what I was missing. I read through the link and (maybe I missed it but) I couldn't find anything to do with args.

Automation/S3/delete_s3_buckets.py
lbrown51: Good point!

farhadbahrehmandhenry (approved):

PR #177, opened 2021-02-18
Redis cache users/user id/tasks

kieranmiguel:
-Added Redis cache for users/userID/tasks that expires after 20 seconds, and reroutes subsequent requests within those 20 seconds to the Cache

-Fixes for Performance Script


functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -39,7 +41,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
...
     #Redis Sever
     try:
-        rDB = redis.Redis(host='localhost',port=5000,db=0) 
+        REDIS_HOST = os.environ['ENV_REDIS_HOST']
+        REDIS_KEY = os.environ['ENV_REDIS_KEY']
+        REDIS_PORT = os.environ['ENV_REDIS_PORT']
+        rDB = redis.Redis(host='ENV_REDIS_HOST', port='ENV_REDIS_KEY', db=0, password='ENV_REDIS_KEY', ssl=True) 
         rDB.ping()
         logging.debug("Connected to Redis!")
     except(redis.exceptions.ConnectionError, ConnectionRefusedError) as e:
```
chonger878, line 50: For line 51, shouldn't it be arg, not arge?

chonger878 (approved):
Just perhaps correct that little "misspelling"  on line 51 of the _init_.py file and thing should be good!

PR #184, opened 2021-02-19
Redis Cache users/userID/tasks and Perf Test Fixes

### LINKED ISSUE #134: [Task] Create Redis cache for users/{userID}/tasks
We can optimize how quickly our app can deal with multiple identical requests by making a Redis cache that will temporarily hold data from previous requests, preventing need to access the SQL server multiple times within a small period of time. 

A Redis cache that holds Read requests for 15 seconds for users/{userID}/tasks, and will divert future read requests for those 15 seconds

MongoDB, ArangoDB, and other virtual Databases would also fit here.

kieranmiguel:
-Added Redis cache for users/userID/tasks that expires after 20 seconds, and reroutes subsequent requests within those 20 seconds to the Cache 

-Redis cache is initialized along with SQL sever, and attempts to connect with try/catch implementation and logging.
-When a get request is made, all tasks are loaded into a JSON object, and then loaded into the redis server as users:user_id:tasks:all, immediately after which, an expire tag is added to the cache, to expire users:user_id:tasks:all after 20 seconds. If another get request is made within those 20 seconds, the request will get from the redis cache instead of the SQL server, reducing lag. 

-Fixes for Performance Script

### DIFF
--- a/functions/HttpTriggerAPIUsersIdTask/__init__.py
+++ b/functions/HttpTriggerAPIUsersIdTask/__init__.py
@@ -4,6 +4,8 @@ import logging
 import os
 import azure.functions as func
 import datetime
+import redis 
+
 
 
 # to handle datetime with JSON
@@ -25,6 +27,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
     method = req.method
     user_id = req.route_params.get('userId')
 
+    
+
+
+
     try:
         conn = connect_to_db()
     except (pyodbc.DatabaseError, pyodbc.InterfaceError) as e:
@@ -32,6 +38,16 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         logging.debug("Error: " + e.args[1])
         return func.HttpResponse(status_code=500)
     logging.debug("Connected to DB successfuly!")
+
+    #Redis Sever
+    try:
+        
+        rDB = redis.Redis(host='nsc-redis-dev-usw2-thursday.redis.cache.windows.net', port='6380', db=0, password='${{ secrets.ENV_REDIS_KEY }}', ssl=True) 
+        rDB.ping()
+        logging.debug("Connected to Redis!")
+    except(redis.exceptions.ConnectionError, ConnectionRefusedError) as e:
+            logging.error("Redis connection error!" + e.arge[0])
+            
 
     try:
         if method == "GET":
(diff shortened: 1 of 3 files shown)


functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -33,6 +39,16 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         return func.HttpResponse(status_code=500)
     logging.debug("Connected to DB successfuly!")
 
+    #Redis Sever
+    try:
+        
+        rDB = redis.Redis(host='nsc-redis-dev-usw2-thursday.redis.cache.windows.net', port='6380', db=0, password='${{ secrets.ENV_REDIS_KEY }}', ssl=True) 
+        rDB.ping()
```
chonger878, line 46: Hm, I am guessing this lets user know somehow that they were/were not able to get onto the Redis DB?

chonger878 (approved):
Wow, I forgot that the further you do down the url, the more complex the code has to be.

jaguiniga (approved):
Looks like you added secrets for keep the password secure. Also the correct Host.

functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -33,6 +39,16 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         return func.HttpResponse(status_code=500)
     logging.debug("Connected to DB successfuly!")
 
+    #Redis Sever
+    try:
+        
+        rDB = redis.Redis(host='nsc-redis-dev-usw2-thursday.redis.cache.windows.net', port='6380', db=0, password='${{ secrets.ENV_REDIS_KEY }}', ssl=True)
```
toddysm, line 45: What if the Redis server host changes?

functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -86,11 +105,24 @@ def get_user_tasks(conn, userId):
         tasks = []
         columns = [column[0] for column in cursor.description]
 
+    if(rDB.get('UsersIdTasks')):
+        unpacked_tasks = json.loads(rDB.get('tasks'))
```
toddysm, line 109: The tasks for which user are you pulling here?

functions/HttpTriggerAPIUsersIdTask/__init__.py
kieranmiguel: Ill add this in the next sprint...

PR #211, opened 2021-02-26
fix the dploy-ui action bug

### LINKED ISSUE #210: fix the deploy-ui action bug
fix the deploy-ui action which is failing at the build step.

farhadbahrehmandhenry:
this PR will fix the bug that causes deploy-ui action to fail.


.github/workflows/deploy-ui.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: |
           cd ${{env.AZURE_WEBAPP_PACKAGE_PATH}}
           npm install
-          npm run build
+          CI='' npm run build
```
theKunte, line 25: What does this do?

farhadbahrehmandhenry: this fixes the bug- the action is failing because of "Treating warnings as errors because process.env.CI = true." and adding this fixes that.

theKunte: Thanks for the clarification! Yes, it does fix it!

theKunte (approved):

jleeboyd (approved):

PR #218, opened 2021-02-28
Redis Expire on get + rDB bug + redisfeatureflag

### LINKED ISSUE #188: [Task] Invalidate Redis Cache for users/userID/tasks on POST, Add flag to disable cache
Currently, my cache invalidates after 20 seconds, I would like the cache to instead, invalidate on a POST to promote use of 
this resource 

Every post request will also invalidate the cache. Add feature flag to turn off Cache.

I was also thinking about update and post requests from other functions, for example, if a task were to be updated, from users/userID/tasks/{taskID} it should also invalidate the cache

kieranmiguel:
Redis Cache now expires on get requests, bug regarding the rDB not being passed to add_tasks method is fixed, redisfeatureflag now exists as CACHE_TOGGLE. 

Whenever a new post request is called, the cache is voided, regardless of if there is data within or not. This method is only called however, when the redis feature flag is enabled either by local.settings.json for testing, or in the azure key-vault during deployment.


functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -40,26 +41,26 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
...
-        rDB = redis.Redis(host='nsc-redis-dev-usw2-thursday.redis.cache.windows.net', port='6380', db=0, password= os.environ["ENV_REDIS_KEY"], ssl=True) 
-        rDB.ping()
-        logging.debug("Connected to Redis!")
-    except(redis.exceptions.ConnectionError, ConnectionRefusedError) as e:
-            logging.error("Redis connection error!" + e.args[0])
-            
 
+    if(redisFeature):
+        try:
+            rDB = redis.Redis(host='nsc-redis-dev-usw2-thursday.redis.cache.windows.net', port='6380', db=0, password='${{ secrets.ENV_REDIS_KEY }}', ssl=True)
```
zxkevinxz: the password needs to be os.environ["ENV_REDIS_KEY"], '${{ secrets.ENV_REDIS_KEY }}' is a reference to the github repo secrets.

zxkevinxz (changes requested):
Redis password line needs to be changed back.

functions/HttpTriggerAPIUsersIdTask/__init__.py
```diff
@@ -159,4 +170,5 @@ def add_tasks(conn, task_req_body, user_id):
         task_id = cursor.fetchval()
         logging.info(task_id)
         logging.debug("task with id {} added!".format(task_id))
+        
         return func.HttpResponse(json.dumps({task_id}, default=default), status_code=200, mimetype="application/json")
```
farhadbahrehmandhenry, line 174: currently, it's not returning taskId in a correct format. should be =>
return func.HttpResponse(json.dumps({"taskId": task_id}), status_code=200, mimetype="application/json")

jaguiniga (approved):
The hard coded information was taken out and it connects to docker server and gets an the array when connected to the SQl DB

PR #225, opened 2021-03-04
Adds React UI for creating a task

### LINKED ISSUE #193: [Task] React UI for creating a task
This task creates the form to allow the UI on our task app to take the form data except the User ID and create corresponding JSON (using JS code) and pass this JSON as a body of the request. The request to the API is created by adding the user Id in the URL for the API.

andluc21:
resolve #193 

This feature adds a form that allows the creation of a new task based on a given user. This added feature takes the form data except the User ID and creates corresponding JSON (using JS code) and passes this JSON as a body of the request to the API endpoint. The userId is implemented into the link - it is a dynamic of the API URL Endpoint of the API that is handled in CreateTask.js

## Testing

- Navigate to the form by pressing the Create Task button on the home page
- Fill out the form with the option of leaving the last two date fields null
- Submit the task and if the operation is successful you will see the notification that it was a success 

Screen Shot 2021-03-03 at 8 02 20 PM

- To change the userId simply go to the CreateTask.js page and swamp out the link with the chosen task link from your development.

Screen Shot 2021-03-03 at 8 16 14 PM

- The goal was to post to this API endpoint `https://nsc-fun-dev-usw2-thursday.azurewebsites.net/api/users/1/tasks` but I believe there is possibly a bug with this link as any number the user input is, the same two users show on the screen. Eventually this link should have create user input functionality  for the userId if this is further developed.

**Troubleshooting Tips**

There are 2 outputs from the console when creating a task, The first number is the userID that the task is going towards and the second field is the task in JSON  format. If these two dont show up it can be linked back to this functions and if a error shows up when creating the task there is a number that will indicate where the error should be.

Date | Activity | Time Spent
-- | -- | --
2/25/2021|Research form handling in React |3 hours
2/28/2021|Create Class version of the form  |2 hours
3/1/2021|Create a Functional version of the form | 2 hours 
3/2/2021|Implement userID function with POST | 3 hours 
3/3/2021|Fix bugs | 4 hours


unicornsleuth:
* closes #193

andluc21:
> * closes #193

zxkevinxz (commented):
Shouldn't the taskId be created when it is entered into the database and not by the user?
Also, I am a little confused if this is supposed to be associated with a user? or if that wasn't a part of this task.

andluc21:
Its a good question! I was kind of thrown for a loop with this task because the api endpoint for the tasks isnt  working completely so this PR creates a resource and the task complete is a 201 resource creation message. It's set up to make a POST and should be going to the endpoint when its hooked up properly. Toddy recommended I make the dynamic part of the URL of the API handled directly in the JS so I just focused on that, I suppose this PR is sort of dependent on another task and might need some updating after that.

PR #244, opened 2021-03-05
Pagination for Users + Users/UserID/Tasks Environ fix 

### LINKED ISSUE #243: [Task] Add Pagination for /users Page
We don't want to load all users for every single get-request, as it would be a waste of resources.

Divide list of users into pages, and only query users from page that is selected, as well as only loading those users into a redis cache entry unique to those users. 

We could treat the redis cache as its own database, and just add, update and delete users from the cache, instead of expiring

kieranmiguel:
Allows for pagination variables to adjust the count and page of users that is selected. 
Format: /users/?count={count}&page={page}

When a request is made, the API "Divides" up the tasks based on the count of users that is selected. For example, is one queries for 10 count and page 5, and there are 50 tasks, tasks 40-50 will appear. If one queries for count 4 and page 5, tasks 16-20 will appear. 

Redis Caching works by setting an individual key for each pagination and count variable. So even queries that contain some of the same tasks will not pull from the same redis key. Redis caches are set to expire after a time of 1200 seconds. This means caching currently only supports identical requests. 

Testing:
To test, send a postman get request in this format: /users/?count=4&page=5 and make sure CACHE-TOGGLE = "On" in local.settings.json

The first request should query the cache and then return from the database,
a second subsequent request should query the cache. This information should be available in the terminal


functions/HttpTriggerAPIUsers/__init__.py
```diff
@@ -3,12 +3,21 @@
...
+from urllib.parse import parse_qs
 import azure.functions as func
 
 USERS_CACHE_KEY = b'users:all'
 CACHE_TOGGLE = os.environ["CACHE_TOGGLE"]
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    url = urlparse.urlparse(req.url)
+    count = parse_qs(url.query)['count']
+    page = parse_qs(url.query)['page']
```
farhadbahrehmandhenry: - where are you using count and page variables. 
- also, I think that would be a good idea to use an if statement to avoid crashing the app when count and page are not defined in url.

farhadbahrehmandhenry (commented):
HttpTriggerAPIUsersIdTask is not working at the moment. replacing line 126 with the following code should fix the problem.
json_tasks = json.dumps(tasks, default=default);

chonger878 (approved):
The documentation placed strategically on the code helps me understand what happens at each point of the program.  A slight suggestion for the task/PR page, perhaps separate the part about count 4 and page 5 or anything to do with queries, give it its' own paragraph, and name that as testing documentation or something...?

gillilands19:
Working for me. And it looks like you took care of the suggestions by Farhad.

gillilands19 (approved):
Working for me. And it looks like you took care of all the suggestions from Farhad.

PR #269, opened 2021-03-14
As/swagger documentation

### LINKED ISSUE #266: Write Swagger documentation for API
This issue creates the Swagger API documentation for our Users/Tasks application.

The easiest to test the code is to plug the code into the Online Swagger Editor, which will create a graphic representation of the API and will spot any errors.

To Test:
1. Copy the code from `/docs/swagger.yaml` 
2. Paste the code into the online [Swagger Editor](https://editor.swagger.io/)
3. The right hand side of the screen will populate with details from my provided code.

Activity | Time Spent
-------------     | ------------- 
Swagger research   | 1.5 hr
Testing endpoints and noting responses    | 1 hr
First draft | 2.5 hr
Revisions | 2 hr
Documentation | .5 hr

AlexSuth:


AlexSuth:
Version 1 of my Swagger API Documentation. Includes all the paths and I think I got all the http methods that are required. I am a little iffy on the http responses, though.

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+          description: "Internal server error"
+    post:
+      tags:
+      - "users"
+      summary: "Add a new user to the database"
+      description: ""
+      operationId: "addUser"
+      consumes:
+      - "application/json"
+      - "application/xml"
```
theKunte: Hi, do we want to include xml applications or only json?

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+  termsOfService: "http://swagger.io/terms/"
+  contact:
+    email: "apiteam@swagger.io"
+  license:
+    name: "Apache 2.0"
+    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
+host: "nsc-fun-dev-usw2-thursday.azurewebsites.net"
+
+schemes:
+- "https"
```
theKunte, line 15: for security should we only use "https" schmes?

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+        type: "string"
+      - in: "body"
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
+      responses:
+        "200":
+          description: "successful operation"
```
theKunte: To be more descriptive you could add something like "User has been updated"

docs/swagger.yaml
AlexSuth: Hey Jenny, xml was included by default, but I think you are right - just json makes sense.

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+      - name: "user_id"
+        in: "path"
+        description: "The name that needs to be deleted"
+        required: true
+        type: "string"
+      responses:
+        "200":
+          description: "User Deleted"
+        "400":
+          description: "Delete Failed"
```
theKunte, line 157: Maybe add a "404": if the user was not found? but 400 would include it too.

theKunte: Also I think the function for "users/{user_id}" includes a patch method and would need to be included ( if a user updates something)

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+      description: ""
+      operationId: "getUsers"
+      produces:
+      - "application/xml"
+      - "application/json"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/User"
```
unicornsleuth: Isn't there a specific way to denote that this will be an array of Users objects?

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+      operationId: "addUser"
+      consumes:
+      - "application/json"
+      - "application/xml"
+      produces:
+      - "application/xml"
+      - "application/json"
+      parameters:
+      - in: "body"
+        name: "body"
```
unicornsleuth, line 48: Isn't this firstName, lastName, and email?

unicornsleuth: Oh wait nevermind, that's in the User object :)

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+        required: true
+        type: 'integer'
+      operationId: "getTasks"
+      produces:
+      - "application/xml"
+      - "application/json"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
```
unicornsleuth, line 179: same as before, show it's an array of Tasks?

docs/swagger.yaml
unicornsleuth: It does have a patch!

docs/swagger.yaml
```diff
@@ -0,0 +1,329 @@
...
+        schema:
+          $ref: "#/definitions/User"
+      responses:
+        "200":
+          description: "Successful operation"
+        "404":
+          description: "User not found"
+        "500":
+          description: "Internal server error"
+    delete:
```
unicornsleuth, line 310: this one also has a patch

unicornsleuth (commented):
Only saw a couple things that looked amiss - otherwise really good!

AlexSuth (commented):
I think I got everything.

theKunte (approved):
changes have been addressed. Looks good!

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        name: "body"
+        description: "User object that will be submitted to create new user"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/User"
```
unicornsleuth: This actually only returns the integer of the new userId

unicornsleuth: { "userId": 1759 }

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
+      responses:
+        "200":
+          description: "User successfully updated"
+          schema:
+            $ref: "#/definitions/User"
```
unicornsleuth: This only returns the string "User updated"

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
```
unicornsleuth: returns the string "user updated successfully"

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+      produces:
+      - "application/json"
+      parameters:
+      - name: "user_id"
+        in: "path"
+        description: "The name that needs to be deleted"
+        required: true
+        type: "string"
+      responses:
+        "200":
```
unicornsleuth, line 154: returns string "User deleted"

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        name: "user_id"
+        description: "The user that needs to be fetched. Use user1 for testing. "
+        required: true
+        type: 'integer'
+      - in: "body"
+        name: "body"
+        description: "Task object that needs to be added to the store"
+        required: true
+        schema:
+          $ref: "#/definitions/Task"
```
unicornsleuth: only required fields for task body are title and description

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/Task"
```
unicornsleuth: no return - returns null

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        in: "path"
+        description: "user that need to be updated"
+        required: true
+        type: "string"
+      - in: "body"
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
```
unicornsleuth, line 104: only required input fields: firstName, lastName, email 
maybe make another object like CreateUser with only the 3 fields?

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+      - name: "task_id"
+        in: "path"
+        description: "The task id that needs to be fetched."
+        required: true
+        type: "integer"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/User"
```
unicornsleuth, line 241: returns a Task object (userId, user, taskId, title, description, createdDate, dueDate, completed, completedDate)

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        in: "path"
+        description: "The task id that will be updated."
+        required: true
+        type: "integer"
+      - in: "body"
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
```
unicornsleuth, line 270: requires a Task object, with title, description, dueDate (dd/mm/yy hh:mm:ss), completed ("0" or "1"), and completedDate (same)
I'm going to file a bug because I don't think all 5 fields should be required for a PUT, but this is how it is at the moment

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+      responses:
+        "200":
+          description: "Successful operation"
```
unicornsleuth: this actually returns just the value "1"

docs/swagger.yaml
unicornsleuth: Actually the PATCH is more lenient, it take any field I think

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        in: "path"
+        description: "The task id that will be updated."
+        required: true
+        type: "integer"
+      - in: "body"
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
```
unicornsleuth: takes a Task object, and add a note that the fields are all optional - it will take whatever fields you pass in

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+      responses:
+        "200":
+          description: "Successful operation"
```
unicornsleuth: returns the value "1"

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        required: true
+        type: "string"
+      - name: "task_id"
+        in: "path"
+        description: "The id of the task that will be deleted"
+        required: true
+        type: "integer"
+      responses:
+        "200":
+          description: "Successful operation"
```
unicornsleuth: returns value "1"

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        type: "integer"
+        format: "int32"
+      createdDate:
+        type: "string"
+        format: "date-time"
+      dueDate:
+        type: "string"
+        format: "date-time"
+      completed:
+        type: "integer"
```
unicornsleuth: I think this actually needs to be a string? When I tested this with "completed": 0, I got a failure - it only worked with "completed": "0"

unicornsleuth (commented):
Ok so I went through Postman and tested these endpoints and updated what the actual responses are!

unicornsleuth:
I'm not sure what the convention is for objects that are different in their input than their output - like, the Task contains references to the userId, user, taskId, but none of those are supposed to be input - they're generated automatically. 
My idea (not sure if it's the right solution) is to have alternative Objects that are CreateUser and CreateTask, and then you can clarify in descriptions that the fields are optional in PATCH but required in PUT.
And use User and Task for the return values.

docs/swagger.yaml
unicornsleuth: Or it would be easy to just add "required fields: firstName, lastName, email" to the description

docs/swagger.yaml
unicornsleuth: You should get rid of lines 108 & 109 here - it isn't returning a JSON object

docs/swagger.yaml
```diff
@@ -0,0 +1,381 @@
...
+        in: "path"
+        description: "user that need to be updated"
+        required: true
+        type: "string"
+      - in: "body"
+        name: "body"
+        description: "Updated user object"
+        required: true
+        schema:
+          $ref: "#/definitions/User"
```
unicornsleuth, line 131: Maybe add to description, "fields optional - only fields provided will be updated"

docs/swagger.yaml
unicornsleuth: & remove lines 137/138 here, as it is not returning a JSON object

docs/swagger.yaml
unicornsleuth: so change line 245 to say /Task

docs/swagger.yaml
```diff
@@ -0,0 +1,397 @@
...
+        name: "body"
+        description: "User object that will be submitted to create new user"
+        required: true
+        schema:
+          $ref: "{\"userId}\": 1759"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/CreateUser"
```
unicornsleuth, line 57: I think you need to switch lines 51-52 with 56-57 here - it takes a CreateUser object as input (same as your CreateTask - looks great!) and then returns whatever the newest userId is - 1759 was just an example, maybe you could just write "int" or something

AlexSuth (commented):
Ready for merge?

unicornsleuth (approved):
Yes! Thanks : )

docs/swagger.yaml
unicornsleuth: Oh I didn't realize these comments didn't go through when I wrote them last night D:

unicornsleuth: Also it should be something like "{\"userId\": int }"

PR #280, opened 2021-03-18
Switch for turning caching on/off for /users/{userid}

### LINKED ISSUE #249: Feature switch to turn on and off caching for /users/{user_id}
**WHAT IT DOES:**
Feature switch to turn on and off caching for /users/{user_id}

**TIME SPENT:**
Date | Description | Time
-- | -- | --
3/9 | Meet with Johart | 1.5 hrs
3/12 | Review class lecture | 1 hrs
3/12 | Code Review and Development | 2 hrs
3/15 | Troubleshooting debugger issues | 1 hrs
3/18 | Debugging | 1 hrs
3/5- 3/18 | Record keeping | 1 hrs

AllisonBC:
Closes #249 
**WHAT IT DOES:**
Feature switch to turn on and off caching for /users/{user_id}

**TO TEST:**
**Set up**
1. _Be sure that you are able to use the VS Code debugger._ Troubleshooting this alone was both difficult and time consuming.
2. Install the Azure Cache extension in VS Code.
image

3. `functions\local.settings.json` file set up:
    a. Need values: "ENV_DATABASE_CONNECTION_STRING", "ENV_REDIS_KEY", "ENV_REDIS_HOST", "ENV_REDIS_PORT"
    b. Add value: `"CACHE_TOGGLE": "true"`
4. Postman is helpful for making the requests.

**Testing**

5. Run the debugger on Python Functions.

6. In Postman or browser make a get request for user with id 1
Terminal:
image
VS Code Azure extension in Caches:
(click on users:all)
image

7. Return to `functions\local.settings.json` file, set `"CACHE_TOGGLE": "false"` and save.
NOTE: At this point, I had to exit and reopen VSCode entirely to get the debugger to acknowledge the change in this value. 

8. Run the debugger on Python Functions.

9. In Postman or browser make a get request for user with id 10
Terminal:
image

VS Code Azure extension in Caches:
image

Running this process a third time with the toggle set to "true" will result in a cache clear, and the user you requested being added to the cache.

**TIME SPENT:**
Date | Description | Time
-- | -- | --
3/9 | Meet with Johart | 1.5 hrs
3/12 | Review class lecture | 1 hrs
3/12 | Code Review and Development | 2 hrs
3/15 | Troubleshooting debugger issues | 1 hrs
3/18 | Debugging | 1 hrs
3/5- 3/18 | Record keeping | 1 hrs


unicornsleuth:
The key should be "users:{userId}" - don't use the universal key - that might be a leftover from other code. You want to build the key each time you cache, by passing in the userId, so you can make something this `redis_key = "users:" + str(userId)` and use that to set/fetch the catch

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -5,8 +5,9 @@
 import json
 import redis
 
-#Global value to be used to invalidate GET,PUT, and DELETE for Redis Cache
+#Global value to be used to invalidate GET, PUT, and DELETE for Redis Cache
 ALL_USERS_KEY = b'users:all'
```
unicornsleuth, line 9: You don't need to use this - if you _want_, you can just save this as b'users:' and then concatenate each userId on it when you query or save to redis.

functions/HttpTriggerAPIUsersId/__init__.py
```diff
@@ -160,25 +161,32 @@ def get_user(conn, user_id, _redis):
 
 
 def get_user_cache(_redis):
-    try:
-        cache = _redis.get(ALL_USERS_KEY)
-        return cache
-    except TypeError as e:
-        logging.critical('Failed to fetch user from cache: ' + e.args[1])
-        return None
+    if CACHE_TOGGLE == "true":
```
unicornsleuth, line 164: I would check with how this variable is set in Azure, and with the other functions. My (human) memory is that the variable was to be "On" or "Off" not "true" and "false." So I save my variable as something like `CACHE_TOGGLE = os.environ['ENV_CACHE_TOGGLE'] == "On"`

unicornsleuth (commented):
Blocking: the redis key for each of these must be "users:{userId}" not "users:all" 
Requested: verify the CACHE_TOGGLE key. Your code works and is clean & easy to read, but my PRs for the cache toggle on /users and /users/{userId}/tasks have already been merged, so they won't work together.

unicornsleuth:
Ok so I just saw that Johart's PR already went through with his CACHE_TOGGLE values as true/false, and then I just checked in Azure and it's been changed to true now, so I guess I was just out of the loop when that got changed and I'll have to make a PR to change my code. Disregard my comments on that note!

functions/HttpTriggerAPIUsersId/__init__.py
AllisonBC: @unicornsleuth   Weird! I wonder why someone changed it without saying anything. I actually saw your code had it set to 'on' a couple weeks ago and mentioned the difference to Johart. I went with what was in Azure because I couldn't figure out how yours worked.

functions/HttpTriggerAPIUsersId/__init__.py
AllisonBC: This doesn't originate from my contribution to the code, but I can look into it

functions/HttpTriggerAPIUsersId/__init__.py
unicornsleuth: Yeah, I don't know when it was changed in Azure - in the lecture where he creates it in class, my memory is that he decided to make it On/Off. But I've got a PR up to change mine now to match yours!

functions/HttpTriggerAPIUsersId/__init__.py
unicornsleuth: Check out the way I implemented this in the UsersIdTask file - at the beginning of the request, I create the redis_key string like this: `redis_key = "users:" + str(userId)` and then pass it into your methods so they can use it when querying or clearing the cache.

AllisonBC: I created a bug issue (https://github.com/North-Seattle-College/ad440-winter2021-thursday-repo/issues/299) since this error was not due to changes from this PR.

functions/HttpTriggerAPIUsersId/__init__.py
unicornsleuth, line 164: If you just remove the `== "true"` this will work better - sometimes it is True, sometimes true, in the Azure function app, but both of those will evaluate to True if you don't tell it to compare Strings. If that makes sense. So everywhere you have `if CACHE_TOGGLE == "true":` to just `if CACHE_TOGGLE:`

unicornsleuth (commented):
There are bugs here - I'm not sure exactly what it is, but when the cache is set to True, it still says caching is set to false in the output. I stepped through it and it looked like it was getting a None result and throwing errors... It could have to do with the thing I pointed out, but I'm not sure.

PR #303, opened 2021-03-23
/Users API returning tasks instead of users, querying redis cache incorrectly 

### LINKED ISSUE #302: [Bug] /Users pulling tasks, pagination not working correctly on /users
**Describe the bug**
/Users page is producing tasks due to a bad sql query, pagination not working correctly on subsequent requests because redis cache key is not being correctly updated 
**To Reproduce**
1. Query the API mutiple times with cache enabled for bad paginations and tasks instead of users 
3. Query API one time for tasks instead of users 

Expected:
Different pagination queries return different amounts of users, not tasks

kieranmiguel:
When a user queries the /Users API, it was returning tasks instead of users. Furthermore, the same redis key was being used for each caching operation, resulting in the same tasks being returned even with changes in pagination.


andluc21 (approved):
Good catch!

zxkevinxz (approved):
bug fix

PR #321, opened 2021-03-25
Fixed PUT bug and Delete Cache

### LINKED ISSUE #288: /taskId PUT failure
**Describe the bug**
A PUT request fails to update, because of a bug in the code - expects 3 positional arguments, but receives 4
After I quickly/temporarily made this fix, I found that a PUT (and also patch, delete) does not invalidate the particular cache for that task.

**To Reproduce**
1. In VSCode, run the Azure functions
2. In Postman, run a GET on http://localhost:7071/api/users/1/tasks
3. Take the first taskId you see (for me, it was 2) and run a GET on http://localhost:7071/users/1/tasks/2 
4. Create a PUT request on http://localhost:7071/users/1/tasks/2 with the raw body:
```
{
   "title": "title",
   "description": "description",
   "dueDate": "03/04/21 12:12:12",
   "completed": "0",
   "completedDate": "04/04/21 12:12:12"
}
```
5. See error

**Expected behavior**
You should see "Error: update() takes 3 positional arguments but 4 were given"

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - VSCode
 - Postman

**To Reproduce 2nd Part**
1. You can easily fix this bug by removing the ",r" from line 308 of the /taskId init file
2. Save & reload functions
3. Run the PUT request from before (step 4)
4. Run the GET request from before (step 3) and see that the data has not been updated
5. Check in VSCode, and see that the output says "Cache cleared" AND "Returned data from cache"
6. Run the GET request from before (step 2) and see that the data has, in fact, been updated. 

**Additional context**
We should have a cache for each task that has been queried, of the form "users:{userId}:tasks:{taskId}"
The caches *are* being created for these, but they are not being cleared when PUT/PATCH/DELETE is run on the task.
The code needs to be reworked to invalidate those particular caches *as well as* the "users:{userId}:tasks" cache on each update.

jaguiniga:
Fixed PUT bug and Delete Cache

### DIFF
--- a/functions/HttpTriggerAPIUsersIdTaskId/__init__.py
+++ b/functions/HttpTriggerAPIUsersIdTaskId/__init__.py
@@ -311,6 +311,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         #if PATCH method is selected, it executes here
         elif method == "PATCH":
             logging.debug('Passed PATCH method')   
+
+            # Invalidate userId tasksId call 
+            invalidate_users_tasks_id(r, userId, taskId)
+            
             return (patch(userId, taskId, task_fields))
         else:
             logging.warn(f"Request with method {method} is not allowed for this endpoint")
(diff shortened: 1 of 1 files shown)


kieranmiguel (approved):

unicornsleuth (commented):
This **does** work for DELETE and PUT, but **not** for PATCH

unicornsleuth (approved):
Works great!

PR #323, opened 2021-03-25
Added function to count how long it takes to load a page 40 times

### LINKED ISSUE #293: [Task]Capture the time to load the users/{user_id}/tasks/{tasks _id} 40 times using Puppeteer with caching
This function should capture the time it takes to load this particular endpoint page 40 times.  It will use Puppeteer to capture that.

In addition to downloading Puppeteer, I also wrote a method that will get the page metrics.  That page metrics will be called 40 times.

None. 

**Addition content**
I give credit to  PR #315 for helping me to get this to work.

**HOW TO TEST THE CODE (Copied from my PR...and I used cmd for this one)**
1.) Navigate to test/ui
2.) Before starting, make sure you type npm install to load dependencies/whatever is needed (especially puppeteer, as I learned that the hard way).
3.) Type node capturetimetaskid.js (note: capturetimetaskid.js is the name of the file)
4.) You should be able to see 40 lines of output stating how long it takes to load the page at every iteration. When the program stops running, you should be able to see the average time it took to load the page 40 times.
5.) IMPORTANT: Please leave the caching disabled, as we will be using redis caching for that.

For example, this should be the expected output for user_id #2, task_id#1:
Screenshot (267)

chonger878:
Closes #293 

**What this PR does:**
This function will not only count how long it will take for a page at the endpoint user/{user_id}/tasks/{task_id} to load, but it will load a page 40 times, counting up the seconds as it goes.  The console will print the average load time.  Caching code is forthcoming

**How to test the code (I used cmd on this one):**
1.) Navigate to test/ui
2.) Before starting, make sure you type npm install to load dependencies/whatever is needed.
3.) Type node capturetimetaskid.js
4.) You should be able to see 40 lines of output stating how long it takes to load the page at every iteration.  When the program stops running, you should be able to see the average time it took to load the page 40 times.

### DIFF
--- a/test/ui/capturetimetaskid.js
+++ b/test/ui/capturetimetaskid.js
@@ -5,6 +5,8 @@ let totalSeconds = 0;
 let loadtime=0;
 
 (async() =>{
+    var startDate = new Date();
+    var startTime = startDate.getTime();
     const ENV_PAGE = "https://nscstrdevusw2thucommon.z5.web.core.windows.net/#/users/2/tasks/1";
     const browser = await puppeteer.launch()
     for(var i = 0; i< NUM_TIMES; i++){
@@ -12,7 +14,9 @@ let loadtime=0;
         console.log(`Opening Page. Heading for target page...`);
         await webpage.goto(ENV_PAGE)
         await webpage.setCacheEnabled(false)
-        loadtime =PerformanceNavigationTiming.loadEventEnd - PerformanceNavigationTiming.loadEventStart;
+            var endDate = new Date();
+        loadtime = endDate.getTime-startTime;
+        console.log(`Time to load page was: ${loadtime}`)
         await webpage.close();    
         }
         console.log(`The average time it took to load the page 40 times: ${totalSeconds/NUM_TIMES}`)
--- a/test/ui/capturetimetaskid.js
+++ b/test/ui/capturetimetaskid.js
@@ -1,8 +1,8 @@
 const puppeteer = require('puppeteer')
 
-NUM_TIMES=40
-totalSeconds = 0
-loadtime=0
+var NUM_TIMES=40
+var totalSeconds = 0
+var loadtime=0
 
 const  pageMetrics = async() =>{
     const browser = await puppeteer.launch({headless: false})
--- a/test/ui/capturetimetaskid.js
+++ b/test/ui/capturetimetaskid.js
@@ -3,9 +3,9 @@ const puppeteer = require('puppeteer')
 var NUM_TIMES=40;
 
 (async() =>{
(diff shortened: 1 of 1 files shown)


test/ui/capturetimetaskid.js
```diff
@@ -0,0 +1,28 @@
+const puppeteer = require('puppeteer')
+
+var NUM_TIMES=40;
+const ENV_PAGE = `https://nscstrdevusw2thucommon.z5.web.core.windows.net/#/users/2/tasks/1`;
+let totalSeconds = 0;
+let loadtime=0;
+
+const  pageMetrics = async() =>{
+    const browser = await puppeteer.launch({headless: false})
```
natni: Launch the browser outside of the for loop to avoid multiple launches. We are required to reload the page not relaunch the browser multiple times.

jaguiniga (approved):

test/ui/capturetimetaskid.js
```diff
@@ -0,0 +1,31 @@
+const puppeteer = require('puppeteer')
+
+var NUM_TIMES=40;
+
+(async() =>{
+    var startDate = new Date();
+    var startTime = startDate.getTime();
+    var totalSeconds = 0;
```
gillilands19: Just a few suggestions. In Javascript variables initialized with the var keyword are hoisted to the top of their scope. This can cause unwanted behavior and bugs. Default to using let and const for variables unless you have a good reason and know why you are using the var keyword.

test/ui/capturetimetaskid.js
```diff
@@ -0,0 +1,31 @@
+const puppeteer = require('puppeteer')
+
+var NUM_TIMES=40;
+
+(async() =>{
+    var startDate = new Date();
```
gillilands19: You are getting the time at the top of the function so you are also timing how long it takes to launch headless chrome along with everything else up to page.goto for the first iteration.

gillilands19 (commented):
The last thing I'll say is that this may give unwanted results, since we are fetching the API data, the initial render is usually a loading page so the page isn't fully rendered yet.. Just a thought for the future.

test/ui/capturetimetaskid.js
chonger878: Ha! I knew it.  "Let" was my first instinct, but then my program crashed and I thought in my panic-filled, dumb brain that the let was crashing it.  Modified it to  "let"

test/ui/capturetimetaskid.js
chonger878: Added {headless:false} to the puppeteer.launch() code.

natni (approved):

